PR #44, opened 2022-06-17
Upper bound circuit

simonmasson:
This PR needs to be merged after PR #41 .
It implements the circuit corresponding to:
* Opening a note commitment with the note data (including `note.value`),
* Checking that `note.value` is less than a given (private) bound.

For that, we modify the note commitment so that it includes more data. It now includes the owner address, the token address, the note value (u32) and the note data (u32). In the long term, we will include all the note data in the commtiment.


cwgoes:
bump @simonmasson what is the status here? should we finish/merge or close this?

XuyangSong:
It seems to be outdated? Some new code has been merged

PR #66, opened 2022-07-29
New layer of curves for efficient hash circuit

simonmasson:
We now use a new layer of elliptic curve so that the Poseidon implementation for the scalar field of the main curve makes sense:
* The implementation in zk-garage works for α=5 and `bls12_377` couldn't use this hash function (it requires `gcd(r-1,5)=1` and it is not the case with this curve).
* Our new layer of curves can use this implementation of Poseidon.
* More over, the Poseidon circuits with this curve are significantly smaller than the ones we could get with `bls12_377`. This layer of curve will probably be integrated to VeriZexe too.


simonmasson:
It requires `cargo update` for the curves and our ark-plonk fork: I modified the custom constrains for `point_addition_gate and `fixed_base_scalar_mult` so that it works for the new curves with `a!=-1`.

I added `PointAddition` and `ScalarMultiplication` gadgets and VPs for tests (and for VampIR examples).

XuyangSong (approved):
lgtm
Could we move the point_add and scalar multiplication tests to ark-plonk? The vp proof generation test is a little heavy. ^^

simonmasson:
I would like to keep (at least) the field addition for the demonstration. This way I can define a VP without all the structures (users, token, notes) but it can be deleted in the long term, for sure. Let's keep them and disable the tests for the proof generation ;-)

PR #67, opened 2022-08-02
Tx construction and performance 

XuyangSong:
1. Tx construction and test
2. Precompute setup parameters and precompile action and blind circuit to reduce test time
3. Add performance.md


src/transaction.rs
```diff
@@ -1,78 +1,360 @@
...
-    created_notes: Vec<(Note<CP>, EncryptedNote<CP::InnerCurve>)>,
-    vps: &'a Vec<ValidityPredicate<CP>>,
+use crate::error::TaigaError;
+use crate::vp_description::ValidityPredicateDescription;
+use plonk_core::circuit::Circuit;
+use plonk_core::circuit::{verify_proof, VerifierData};
+use plonk_core::proof_system::{pi::PublicInputs, Proof, VerifierKey};
+use rand::RngCore;
+
+pub const NUM_TX_SLICE: usize = NUM_NOTE;
```
simonmasson, line 17: Each transaction slice corresponds to spending one note and creating one note?

simonmasson: If yes, could we consider a note consumption without creating a new one? Maybe it does not make sense ahah

src/transaction.rs
```diff
@@ -1,78 +1,360 @@
...
+
+pub struct Transaction<CP: CircuitParameters> {
```
simonmasson, line 19: A transaction contains:
* one action per {spend+create} corresponding to the binding of the vks, etc.
* one spend check including the sender proof and the token proof
* one output check (is the name accurate?maybe creation_slice?) for the receiver proof and the token proof.

src/transaction.rs
```diff
@@ -1,78 +1,360 @@
...
+    pub vp_public_inputs: PublicInputs<CP::CurveScalarField>,
+    // The vk of vp is blinded partially, blinded parts can be constructed from blinded_vp_public_inputs.
+    // The unblinded_vp_vk is the unblinded parts.
+    // TODO:
+    pub unblinded_vp_vk: VerifierKey<CP::CurveScalarField, CP::CurvePC>,
+    // blind vp proof
+    pub blind_vp_proof: Proof<CP::CurveBaseField, CP::OuterCurvePC>,
+    // The public inputs for blind vp proof
+    pub blinded_vp_public_inputs: PublicInputs<CP::CurveBaseField>,
+    // TODO:
```
simonmasson, line 56: I don't get what it is... but it is a TODO :-D

src/transaction.rs
```diff
@@ -1,78 +1,360 @@
...
+}
+
+impl<CP: CircuitParameters> ActionSlice<CP> {
+    pub fn build(
+        action_public: Action<CP>,
+        action_circuit: &mut ActionCircuit<CP>,
+    ) -> Result<Self, TaigaError> {
+        let setup = CP::get_pc_setup_params(ACTION_CIRCUIT_SIZE);
+        // Compile the circuit
+        let pk_p = CP::get_action_pk();
```
simonmasson, line 87: (useless comment)
For me, "pk" means "prover key" and "vk" means "verifier key" so I guess we should remvoe `_p`.

src/transaction.rs
```diff
@@ -1,78 +1,360 @@
...
+        let vk = CP::get_action_vk();
+
+        // Prover
+        let (action_proof, pi) =
```
simonmasson, line 91: (useless comment)
pi can be confusing and could be replaced with `(action_)public_inputs` instead (in all the repo)

simonmasson (approved):
I only left comments on the transaction part. Other files are mostly refactor of the previous code.

My main remark is that the transaction is split into four steps of "I spend a note and I create a note". @joebebel , is it how we want to implement it?

Apart of it, I would say LGTM :-)

src/transaction.rs
XuyangSong: It's an action slice with one input note and one output note to generate an action proof.
Both input note and output note can be dummy(zero, or invalid note) so that we can construct an action without creating a new one.

src/transaction.rs
XuyangSong: 1. Some public inputs are needed to check between the proofs. I don't want to handle the magic variable indexes now, since our implementation are changing. Let's fix this later.
2. After handling the magic variable indexes, we can replace `PublicInputs<F>` by `Vec<F>`. I think It's more reasonable.
3. Some parameters in VP are need to be designed.

joebebel:
Also looks good to me.
I think it's fine to have each transaction have exactly NUM_NOTES Action. VP's must have fixed size input anyway, and we can handle other cases later.

Issue #72, opened 2022-08-08
Optimization of the blinding circuit


simonmasson:
Our current blinding circuit is decomposed into:
1. Addition and fixed base scalar multiplication for the 6 polynomial selector precomputations,
2. Poseidon hashes for the coordinates of the 6 polynomial selector precomputation,
3. A final Poseidon hash of the previous hashes,
4. (not yet) A blake2 hash.

These blinding circuits are an expensive part of the overall transaction, and it is worth optimizing it. The following optimizations are probably possible:
* 1: Scalar multiplication algorithm can use the GLV optimization available for BW6 curves.
* 2: Our Poseidon hashes have inputs the `x` and `y` coordinates of each points, but `y` is only one bit information. We could probably hash only `x` and use smaller hash circuits. We need to figure out if it is secure to do it this way.

Note that the blinding method will be modified in the long term, but the efficient SM circuit for BW6 curves (and BLS12 as well) would be also useful for the alternative we target.

PR #73, opened 2022-08-08
Optimization of the blinding circuit

XuyangSong:
#72 

- [x] Hash the 6 blinded elements instead of all in vp. An improvement about blind circuit size from 2^15 to 2^13


src/circuit/blinding_circuit.rs
```diff
@@ -74,22 +69,29 @@ where
...
+        //         poseidon_circuit.output_hash(composer)
+        //     })
+        //     .collect::<Vec<Variable>>();
+
+        // poseidon_circuit.reset(composer);
+        // for v in hash_vec.iter() {
+        //     poseidon_circuit.input(*v).unwrap();
+        // }
+
+        // Compress blinded elements in vp
```
simonmasson, line 90:
```suggestion
        // Compress x-coordinate blinded elements in vp
```

src/vp_description.rs
```diff
@@ -50,29 +51,35 @@ impl<CP: CircuitParameters> ValidityPredicateDescription<CP> {
...
-                poseidon.reset(&mut ());
-                for x in hash_vec.iter() {
-                    poseidon.input(*x).unwrap();
-                }
+                // poseidon.reset(&mut ());
+                // for x in hash_vec.iter() {
+                //     poseidon.input(*x).unwrap();
+                // }
+
+                // Compress blinded elements in vp
```
simonmasson, line 79:
```suggestion
                // Compress x-coordinate blinded elements in vp
```

simonmasson (approved):
LGTM. It reduces the (partial) blinding circuit but if we keep this blinding construction (with a tweak for the σ polynomials), we will go back to larger circuits as before.

XuyangSong:
Not needed so far

PR #75, opened 2022-08-15
Update user.rs and users.md

Acentelles:


simonmasson:
This was some code for the italy meeting explanation. It involves almost only the `docs/` directory that is not really relevant for our implementation.
The code splits the key generation into several steps but it is not needed in practice. I think it was more from a pedagogical point of view. It is mostly outdated since it involves only zk-garage PLONK implementation. I don't think the `dummy_of_range` function is needed anymore with the current state of implementation.

PR #76, opened 2022-08-19
Replace el_gamal unit test with prop test

oxarbitrage:
Hello there, taiga looks great and i am trying to learn more on how it works. Tests always help to understand a new codebase so thank you for making those. 

While playing with some tests i found you can have a bit more coverage in `el_gamal_test` if it is "upgraded" to a prop test.

Feel free to discard this PR if this is not the right time to do this or for any other reason.

Thank you again for the hard work.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ rand_core = {version = "0.6", default-features=false, features = ["getrandom"] }
 
 [dev-dependencies]
 ark-std = "0.3"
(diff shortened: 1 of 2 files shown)


cwgoes:
Thank you for the contribution! Alas I think we are no longer using the ElGamal code and we should probably delete it (cc @XuyangSong)

XuyangSong:
Thanks, nice suggestion! we're considering to update to prop test

PR #81, opened 2022-09-07
Exec model description draft

vveiln:
Here is a draft version of the execution model description based on the discussion we had during the retreat


book/src/exec.md
```diff
@@ -0,0 +1,55 @@
+## On a high level
+Users use intents when they want to exchange their assets for some other asset. High level description of the flow:
```
Acentelles, line 2: I'd say bartering/exchange of assets is a specific case of an application

book/src/exec.md
```diff
@@ -0,0 +1,55 @@
...
+3. **Solve**: Solvers find matching intents and update the intent until it is fully satisfied
+4. **Finalize**: When the intent is satisfied, a transaction is created and published on the blockchain
+
+![img_1.png](img/img_exec_high.png)
+
+### Create an intent
+User intents can be seen as partial transactions. Users send them to the intent gossip network and solvers match them in order to create full transactions and publish them on the blockchain. That implies that users need to give some information to the solvers that is sufficient to create the final transaction (proofs). To make sure that users reveal the minimal amount of private infortmation that is still enough to create the transaction, users create **intermediate notes** from the notes they are willing to spend.
+
+#### Intermediate notes
+An intermediate note is a note with a simplified sending VP. This VP is specific to the intent and contains only the rules that are necessary to satisfy for the transaction to happen. All of the note fields stay the same except:
```
Acentelles, line 14: How is the VP simplifed in an intermediate note?

book/src/exec.md
```diff
@@ -0,0 +1,55 @@
...
+- sending VP -> intept-specific VP
+- owner address -> intent-specific owner address (sending VP is used to calculate the owner address, so changing the VP implies the change of the owner address)
```
Acentelles, line 16: Why does the intent-specific VP change?

book/src/exec.md
```diff
@@ -0,0 +1,55 @@
...
+Note: The content of intent-specific VPs in known to the solver because the solver will create the final transaction and some of the proofs
+
+Note: intermediate notes will not be published
+
+### Solve
+We are considering the model when a solver makes one step at a time and sends the result to the next solver. In practice, the solver can send the result to themselves and continue solving if they have the intent to make the next step. It would be nice to merge the steps into one when possible, but for simplicity we ignore this detail here
+
+Note: each solver must check the proofs they receive from other nodes
+
+When a solver has two intents that can be matched together, they match the intents by spending the old notes and creating new notes. Solvers have the authority to spend and create the intermediate notes they receive and produce the proofs required to perform the action. Solvers also know the content of intent-specific VPs and receving VPs of the users (necessary to be able to satisfy them). Solvers don't know the identities of the users. Created notes are added in a local MT tree that will be published later along with the settled transaction.
```
Acentelles, line 31: If solvers can spend and create, they need to be a participant in the blockchain, right?

Does the solver need to be identified somehow to be trusted by users?

book/src/exec.md
```diff
@@ -0,0 +1,55 @@
+## On a high level
+Users use intents when they want to exchange their assets for some other asset. High level description of the flow:
+1. **Create**: User creates an intent publishing the information of what they have and what they want to get in exchange
+2. **Gossip**: The intent goes to the intent gossip network and gossiped around the solver nodes.
+3. **Solve**: Solvers find matching intents and update the intent until it is fully satisfied
```
cwgoes, line 5: note that 1-3 may cycle until a final transaction is found

book/src/exec.md
cwgoes: yeah. we should use more general language such as "enact a multiparty state transition"

book/src/exec.md
cwgoes: It includes only the rules governing the specific pieces of state in question, and those pieces of state are sent from the user VP to this temporary VP

e.g. if Alice wants to trade A for B, Alice would first send A from her user VP to a temporary VP stipulating that A can be sent to anyone who sends her B

it'd be good to keep a running example through these notes

book/src/exec.md
cwgoes: solvers do not need to be identified since all the computations they're performing are ultimately checked by proofs (or transparently)

these spends and creates may be only spends & creates of intermediate notes, so the solvers don't necessarily need accounts on the chain (though if they want to receive fees they will at least need an address/VP at which to do that)

book/src/exec.md
```diff
@@ -0,0 +1,55 @@
...
+
+If a solver has two intents that can be matched together, two cases are possible:
+1. The match is partial, partial transaction and a new intent are created. In this case the solver appends the data they produced to the new intent and gossips the new intent to the next node
+2. The match is final, the solver publishes the final transaction as described below
+
+Currently we assume a simpler model where only one solver can match n-party bartering intents (no partial solving) and an intermediate tx cannot be published on the blockchain as VPs will not allow that (balance issue). In the partial solving case partial transactions cannot be published on the blockchain as there will be a VP/other restriction mechanism blocking that.
```
cwgoes, line 37: currently, as in, in the current implementation?

cwgoes (changes requested):

book/src/exec.md
vveiln: Yes

book/src/exec.md
vveiln: By 'change' here I mean that going from userVP to intent-specific VP will lead to a different owner address in the new note. I'll rephrase that

book/src/exec.md
vveiln: Don't solvers also create the final, non-intermediate note?

book/src/exec.md
cwgoes: @joebebel how difficult would this be to change?

book/src/exec.md
cwgoes: can we add a diagram here for this flow?

book/src/exec.md
cwgoes: yes - but that doesn't necessarily require them to be a participant in the blockchain. why would it?

joebebel: I agree, in general the solver doesn't receive or hold any state or tokens on chain. Even in the case where the solver might create notes, this would be authorized by some other VP and not the solver's VP or signature.

book/src/exec.md
joebebel: I don't think it'll be very hard, but it's not so clear to me what the best way to do it is. One simple solution is to bring back homomorphic balance commitments and allow token VPs to "opt in". Then for any token that only balances in the usual way, there's no issues with partial solving or private solving, and we can discuss further how to allow it for other tokens.

XuyangSong:
When creating an intermediate note(or intent), do we need to generate an `ActionProof` which proves an old note consumption and a new note creation(the intermediate note)?

As for proofs in partial tx,  I can implement #82 to support the `user send vp` proof generation. And we might not use the `user send vp` which needs to know output notes or other notes in intent construction for now. For token vp proof generation, I'm afraid that it's tricky since the `token vp` needs to know all the notes in the tx under the current architecture . I agree with Joe to bring back the homomorphic commitments to check the balance. Then the `token vp` might not have to know all the notes.

vveiln:
As I understand, we need to produce all of the proofs for partial transactions, including the `ActionProof`

About `tokenVP`: my understanding was that for VPs partial transactions will look like normal transactions, so the notes the VPs can see will be seen as "all of the tx notes" to them, except that the tx is partial

cwgoes:
Can we update this for Joe's latest proposal?

book/src/exec.md
```diff
@@ -0,0 +1,154 @@
+# Taiga execution model
+
+## On a high level
+Intents are a mechanism that allows users to enact a multiparty state transition. High level description of the flow:
+1. **Create**: A user creates an intent publishing the information of what they have and what they want
+2. **Gossip**: The intent goes to the intent gossip network and is gossiped around the solver nodes
+3. **Solve**: Solvers find matching intents and update the intent with partial transactions until the intent is fully satisfied
+4. **Finalize**: When the intent is satisfied, a transaction is created and published on the blockchain
+
+![img.png](img/exec_high.png)
```
Acentelles, line 10: Screenshot 2022-10-24 at 11 54 00

This is how I understand the high level execution model:
In circles we have the parties, arrows represent what is being sent, dotted lines represent actions.

I think it would be helpful to indicate that solvers and users are constantly listening to the gossip network.

From this and your diagram, users don't send intents directly to a solver. Is that correct?

vveiln: I focused on the fact that users send intents and solvers receive intents, and everything in between is out of scope, but maybe it is wrong. Also, I'm not sure that what happens in between is Taiga-specific, but if it is, it definitely should appear somewhere

book/src/exec.md
```diff
@@ -0,0 +1,169 @@
...
+How intent tokens are used:
+1. Spending notes in the initial partial transaction, the user additionally spends an intent token note of value [1].
+The 'twin' intent token note of value [-1] balancing this token will only be created if the user's intent (intentVP) is satisfied.
+3. Once all of the parties involved in a state transition create their [-1] intent token notes (meaning that the intents of all of them are satisfied), the transaction is balanced and can be finalized.
+
+![img.png](img/exec_intent_notes.png)
+
+**Note**: instead of using notes of negative value the user can instead spend a [1] intent token note (that hasn't been created before) in their initial partial transaction. Once the user's intent is satisfied, a [1] intent note will be output, balancing the initial spend.
+
+**Note**: intent token notes are not committed to in the global CMtree
```
XuyangSong, line 54: As Joe [proposed](https://github.com/anoma/taiga/pull/85#issuecomment-1302925442), the intent notes will be committed to the tree. The code has been updated.

It is my understanding that intent notes(or general dummy notes) have no difference with "normal" notes except that intent notes don't need to check the merkle path authorization. Therefore, an intent note will be created as output-note(stored on the tree) in ActionCircuit when a user wants to create an intent token note of value [1]; then the intent note will be consumed as spent-note in ActionCircuit when the intent vp is satisfied and we can create the [-1] intent note. Finally, the sum of output-notes is balanced with the sum of spent-notes. Does it make sense?

book/src/exec.md
```diff
@@ -0,0 +1,169 @@
...
+### Intent tokens
+
+**Intent tokens** are a mechanism that makes sure that the partial transaction is *unbalanced* until the corresponding intent is satisfied, making it impossible to publish the transaction.
+
+Intent token notes *cannot be spent*, only created. Instead of spending a note of such type, intent notes with *negative value* can be created.
+
+How intent tokens are used:
+1. Spending notes in the initial partial transaction, the user additionally spends an intent token note of value [1].
+The 'twin' intent token note of value [-1] balancing this token will only be created if the user's intent (intentVP) is satisfied.
+3. Once all of the parties involved in a state transition create their [-1] intent token notes (meaning that the intents of all of them are satisfied), the transaction is balanced and can be finalized.
```
XuyangSong, line 48: skip to 3 from 1

book/src/exec.md
vveiln: Ah, I didn't see this comment, thanks! Do we actually have negative value notes or we gonna have inverse life cycle (instead of creating [-1] and [1]  note, spend a [1] note first and then output it when the VP is happy)?

XuyangSong: In my current implementation, output an intent note first and spend it when the vp is satisfied.
We can have a discuss to see which is better?

vveiln: Not sure I understand the difference between outputting first and spending first (i'll drop a message on slack). I'll correct the spec to be in sync with the code

book/src/exec.md
```diff
@@ -0,0 +1,168 @@
...
+2. **Gossip**: The intent goes to the intent gossip network and is gossiped around the solver nodes
+3. **Solve**: Solvers find matching intents and update the intent with partial transactions until the intent is fully satisfied
+4. **Finalize**: When the intent is satisfied, a transaction is created and published on the blockchain
+
+![exec_high.png](img/exec_high.png)
+
+## Step 1: Create an intent
+
+Intents specify ephemeral interests of users. The step of intent creation consists of two parts:
+1. **Specify an intent**. `IntentVP` encodes the user interests in the [VP](./validity-predicates.md) form. As in a valid transaction all VPs must evaluate to `True`, the intent is guaranteed to be satisfied.
```
cwgoes, line 17: We should clarify what "intent VP" is - in my understanding intent VP is not a distinction in the code, only a distinction in usage, right?

book/src/exec.md
```diff
@@ -0,0 +1,168 @@
...
+2. **Create the initial *partial transaction***.
+
+### Partial transaction
+
+Two of the main requirements for an Anoma transaction to be valid are: 
+1. VPs of all involved in the transaction parties must be satisfied.
+2. The transaction must be **balanced** , i.e the value spent in the transaction minus the value output equals to some predefined value `v`, the **balancing value**.
```
cwgoes, line 24: This balancing value would be transparent, right? (and maybe involved in interactions with the transparent system)

book/src/exec.md
```diff
@@ -0,0 +1,168 @@
...
+If we look at the intent token notes as messages passing from one user (application) to another, the output [1] intent note can signal a message being sent, and spending of that note will signal that the message has been received. See more about the idea of message passing [here](message_passing.md).
+
+## Step 2: Gossip
+Once the intentVP is specified and the initial partial transaction is created, the user sends everything to the intent gossip network. Gossip nodes send those intents to each other, spreading them around. Solvers listen to the intent gossip network and receive the intents.
+
+## Step 3: Solve
+
+Solvers receive intents from the intent gossip network and match them together in order to create transactions. 
+
+**Note**: that implies that users need to give some information to the solvers (including the content of `intentVP`) that is sufficient to create partial transactions and transactions (to create proofs).
```
cwgoes, line 62: We should be more specific here - in order to match two partial transactions, what does the solver need?
- intent VPs of both partial transactions
- note commitments / nullifiers from prior partial transactions which will be chained
- proofs from prior partial transactions which will be chained

anything else?

book/src/exec.md
```diff
@@ -0,0 +1,168 @@
...
+**Step 1-2**: Alice and Bob create their intentVPs and initial partial transactions. Alice spends both of the notes she is ready to give away, expecting to receive one of them back (the other one will go to the counterparty).
+
+**Step 3**: A solver sees Alice's and Bob's partial transactions and matches them together. Alice receives the blue dolphin NFT and one of her notes ([1] of A) back, her intentVP is satisfied and Alice's intent token note is spent. Bob get's [2]B note from Alice, his intentVP is satisfied, his intent token note is spent. All total per-token balances are equal to 0, and the final transaction can be created.
+
+**Step 4**: The final transaction is created using the spent and output notes from the partial transactions.
+
+### 3. One way to represent arbitrary states
+
+For an arbitrary application, notes store the state of the application. When the application needs to change its state, it (i.e. any party that has the authority) can spend the old state and produce a new state.
+If the state change is possible within one partial transaction, such a note has a value 0 and doesn't affect the total balance.
```
cwgoes, line 153: I'm not sure that I understand the distinction here - is state _not_ consumed in this case? Wouldn't it still be invalid to consume the old note twice?

cwgoes (changes requested):

book/src/exec.md
vveiln: Not a distinction from other VPs or from intents?

book/src/exec.md
vveiln: The old state is consumed (once) in all cases. Zero-value state notes in this case are not dummy, so if the nullifier is published, the note can't be spent twice, as I understand

In both cases (state changed within one transaction or state changed across multiple partial transactions) the old state and the new state notes have zero value, but in the latter case intermediate state notes have non-zero value (to make sure the tx is unbalanced until the final state is computed)

book/src/exec.md
cwgoes: We have send/receive VPs on notes, right? And an intent VP is a usage of the send VP to encode the desires of a user (also receives X...)? (perhaps my understanding is out of date)

book/src/exec.md
cwgoes: ah I see, makes sense

book/src/exec.md
```diff
@@ -0,0 +1,174 @@
+# Taiga execution model
+
+## On a high level
+**Intents** are a mechanism that allows users to enact a multiparty state transition. Users express their interests with intents, and Taiga makes sure that the intents are satisfied atomically.
+
+Here is the high level description of the process, starting from creating intents and ending up publishing a blockchain transaction representing the desired state transition:
+1. **Create**: A user creates an intent, publishing the information of what they have and what they want
+2. **Gossip**: The intent goes to the intent gossip network and is gossiped around the solver nodes
+3. **Solve**: Solvers find matching intents and update the intent with partial transactions until the intent is fully satisfied
```
simonmasson, line 9: By "updating", you mean that with the creation of these partial transactions, the intents may (or may not be satisfied? may not) be satisfied, right?

book/src/exec.md
```diff
@@ -0,0 +1,174 @@
...
+**Intent tokens** are a mechanism that makes sure that the partial transaction is *unbalanced* until the corresponding intent is satisfied, making it impossible to publish the transaction.
+
+Intent token notes are **dummy** notes - meaning that unlike "normal" notes, the merkle path isn't checked for them (but they can have arbitrary value and stored in the CMtree, just like "normal"). 
+
+**Note**: being a dummy note and having zero value are two independent concepts in Taiga. Zero value notes aren't necessarily dummy, dummy notes can have non-zero value.
+
+How intent tokens are used:
+1. Spending notes in the initial partial transaction, the user additionally outputs intent token note of value [1].
+This note will only be spent if the user's intent (intentVP) is satisfied.
+2. Once all of the parties involved in a state transition spent their intent token notes (meaning that the intents of all of them are satisfied), the transaction is balanced and can be finalized.
```
simonmasson, line 50: So iiuc, each user creates a (intent token) note with what they want to send and a boolean (`[1]~true`). Would it be possible to do the same thing without the intent token note?

simonmasson: maybe it is simply an abstraction for the "message passing" explained in the next paragraph?

book/src/exec.md
```diff
@@ -0,0 +1,174 @@
...
+Solvers are responsible for creation of all proofs (`Action`, `VP`, etc) required for their partial solutions. For that reason, solvers:
+- **have the authority** to spend and create the notes they receive and produce the proofs required to perform the action,
+- **know** the content of intentVPs of the users (necessary to be able to satisfy them),
+- **don't know** the identities of the users.
+
+When solvers receive partial solutions from the gossip network, they must check all of the proofs attached to them.
+
+#### Local `cm` trees
+
+To store note commitments of the created notes, a local commitment tree `CMtree` is created. After the transaction is finalized, the tree will be published on the blockchain along with the transaction.
```
simonmasson, line 86: We cannot just consider the final transaction and forget about the partial ones (and so we don't consider the local `CMTree`)?

book/src/exec.md
```diff
@@ -0,0 +1,174 @@
...
+
+Three parties, Alice, Bob, and Charlie, are looking for some assets in exchange for something else. Their intents can be matched into a three-party bartering cycle. Let's see step by step how this happens.
+
+**Note**: parties don't require three-party bartering explicitly.
+
+![img.png](img/exec_3_party.png)
+
+**Step 1-2**: The users define their intents and create intentVPs. It doesn't have to happen at the same time for all users, but for simplicity we describe it as one step. The intentVPs of all three users have the same structure: users are willing to spend their asset (Alice - a star, Bob - a dolphin, Charlie - a tree) in exchange for some other asset. Once the user receives the desired asset, the intent token note is spent. In addition to that, all three users also create their initial partial transactions spending the asset they are ready to give away (Alice ptx, Bob ptx, and Charlie ptx).
+
+**Step 3**: A solver sees Alice's ptx and Bob's ptx, matches them together, and creates a new partial transaction. Alice's intentVP is satisfied, and her intent note is spent.
```
simonmasson, line 118: I am thinking about the general case where a solver receives `n` partial transactions. Given two ptx, the solver can always do this step 3, and so it creates o(n²) ptx. This seems to be a lot, but maybe I didn't get how it works in practice...

vveiln: If a solver sees a ptx sending BTC and expecting NAM and another ptx sending XAN and expecting ETC, it doesn't make sense for the solver to match them

Plus, matching that many ptx together in all possible combinations considering that no more than n/2 of them will be a part of real transactions seems like a waste of time for the solver

book/src/exec.md
vveiln: To create proofs for intermediate ptx (when you spend a note that was created in the previous ptx, it doesn't have `cm` published in the global `CMtree`) we need to have local`CMtree`

book/src/exec.md
vveiln: In a simple example, each user spends a note they want to give away and creates an intent token note of value [1]. Intents tokens are here to make sure the transaction is unbalanced if any of the parties didn't get what they wanted. Once there is another ptx where the user receives whatever they wanted, the user spends this intent token note, balancing the tx

book/src/exec.md
vveiln: Basically it means to add freshly created ptx to the "project". If everyone is happy, then the final tx is created, if not - solvers keep creating and adding ptxs to the "project"

book/src/exec.md
simonmasson: Mhhh I see, I forgot we need to spend the ptx :-D

book/src/exec.md
simonmasson: Yes, that's my point. With this description of solver, it seems that they are waiting their time. From what you said, in practice their have other conditions that refine their search. Thanks!

book/src/exec.md
simonmasson: I see. I was wondering if it is not possible to achieve this without this `[1]`, in order to simplify a bit, but maybe it is not that simple :-D

PR #84, opened 2022-09-28
Simplify psi in note

XuyangSong:
In Orchard `cm = SinsemillaCommit(...|| rho || psi || value, rcm)`, `psi = prf([9] || rho || rseed)` and rcm = `prf([5] || rho || rseed)`.
I've removed the psi derivation from circuit and generated `psi` and `rcm` randomly for now.
Do we need to keep the `prf` derivation for `psi` and `rcm` like Orchard? It seems they can only transfer `rseed` in note-encryption instead of `psi` and `rcm`. Any other consideration?
Further, I didn't figure out why we need the `psi` here. If we remove `psi`, will we have any security problems?

### DIFF
--- a/taiga_halo2/src/action.rs
+++ b/taiga_halo2/src/action.rs
@@ -70,7 +70,7 @@ impl ActionInfo {
         ActionInfo::new(spend_info, output_info)
     }
 
-    pub fn build(self, rng: &mut impl RngCore) -> (ActionInstance, ActionCircuit) {
+    pub fn build<R: RngCore>(self, mut rng: R) -> (ActionInstance, ActionCircuit) {
         let nf = self.spend.note.get_nf();
 
         let user = User {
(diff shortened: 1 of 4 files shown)


joebebel:
@XuyangSong 
Oops, I completely forgot to reply to this, sorry!

> I've removed the psi derivation from circuit and generated psi and rcm randomly for now.

This is fine for now but I think we should probably decide the right way to do it.

> Do we need to keep the prf derivation for psi and rcm like Orchard?

We should have something similar.

> It seems they can only transfer rseed in note-encryption instead of psi and rcm. Any other consideration?

It does save a few bytes in the note encryption, but also, in case rseed is biased a little bit, it could be harder to exploit. In theory only good randomness sources should be used, but ECDSA is quite badly broken because of this. There could also be other reasons.

> Further, I didn't figure out why we need the psi here. If we remove psi, will we have any security problems?

Psi is needed to get statistical privacy of the nullifier (as opposed to assuming that Poseidon is a PRF or the discrete-log assumption) - more information is [here under "use of psi"](https://zcash.github.io/orchard/design/nullifiers.html)

@simonmasson - confusingly, this derivation is for dummy notes only and spending real notes uses a different rho - it is important that this construction is used to avoid the Faerie Gold attack described later in the document

XuyangSong:
@joebebel 
Oh, I see. Very thankful!
I'll add a prf to generate the random psi and rcm later.

PR #87, opened 2022-10-24
application vp reorg

XuyangSong:
- Move `User` into `Application`, and the user data won't be derivated in value base
- Use general user_data instead of user vps. Since the `nk` needs to be binded with the send_address/vp/data, I keep the basic structure of `User = Com( Com(send_data, nk), recv_data)` if it makes sense.

### DIFF
--- a/taiga_halo2/src/action.rs
+++ b/taiga_halo2/src/action.rs
@@ -6,9 +6,6 @@ use crate::{
     net_value_commitment::NetValueCommitment,
     note::{Note, NoteCommitment},
     nullifier::Nullifier,
-    user::User,
-    user::UserSendAddress,
-    vp_description::ValidityPredicateDescription,
 };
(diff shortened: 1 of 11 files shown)


Acentelles (approved):

vveiln:
Do I understand correctly that `Application` now becomes the owner of both `User` and (consequently) the notes that belong to the user?

XuyangSong:
> Do I understand correctly that `Application` now becomes the owner of both `User` and (consequently) the notes that belong to the user?

In my point of view, `Application` has a `User/Address` and defines the format/VP(s) of `User/Address`.

XuyangSong:
merge into branch [xuyang/net_value_commitment](https://github.com/anoma/taiga/tree/xuyang/net_value_commitment)
Let's discuss it under #85

PR #89, opened 2022-11-02
Sudoku example

Acentelles:


simonmasson (commented):
I review only the valid_puzzle circuit, and I am trying to implement the necessary changes in order to get feasible circuits.
In my laptop, it tooks ~99 sec for building the circuit. It should be more efficient for this simple circuit...

taiga_halo2/examples/sudoku/app/valid_sudoku.rs
```diff
@@ -0,0 +1,447 @@
...
+                    .collect();
+                squares.push(square.concat());
+            }
+        }
+
+        // check that rows, cols and small squares are permutations of (1..10)
+        for (cpt, perm) in [rows, cols, squares].concat().iter().enumerate() {
+            // lhs = prod_i (i + gamma)
+            // rhs = -prod_i (sigma(i) + gamma)
+            // compute the circuit corresponding to lhs + rhs (should be zero)
```
XuyangSong, line 262: Are we using `gamma` as a random challenge to check the permutation?

taiga_halo2/examples/sudoku/app/valid_sudoku.rs
```diff
@@ -0,0 +1,447 @@
...
+            .iter()
+            .enumerate()
+            .for_each(|(i, solution_cell)| {
+                let puzzle_cell = assign_free_instance(
+                    layouter.namespace(|| "instance"),
+                    config.primary,
+                    i + 27,
+                    config.advices[0],
+                )
+                .unwrap();
```
XuyangSong, line 338: Do we need to public the puzzle_cell here? not witness the puzzle_cell

taiga_halo2/examples/sudoku/app/valid_sudoku.rs
Acentelles: Yes. See Lemma A.3 [here](https://eprint.iacr.org/archive/2019/953/1656500364.pdf)

XuyangSong: I'm not sure if we can use in this way here. It seems that `gamma` is deterministic?

taiga_halo2/examples/sudoku/app/valid_sudoku.rs
Acentelles: The puzzle is a public input

taiga_halo2/examples/sudoku/app/valid_sudoku.rs
```diff
@@ -0,0 +1,447 @@
...
+        let u: Vec<u8> = s1
+            .iter()
+            .zip(s2.iter())
+            .map(|(b1, b2)| {
+                // Two entries of the sudoku can be seen as [b0,b1,b2,b3] and [c0,c1,c2,c3]
+                // We store [b0,b1,b2,b3,c0,c1,c2,c3] here.
+                assert!(b1 + 16 * b2 < 255);
+                b1 + 16 * b2
+            })
+            .collect();
```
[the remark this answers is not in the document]
simonmasson: Nice catch @XuyangSong ! ;-) This was from an old code I wrote. We should add a zero to `s1` if `sudoku.len()%2 = 1`.

taiga_halo2/examples/sudoku/app/valid_sudoku.rs
simonmasson: It is used as a Fiat-Shamir transform with the result of Lemma A.3 where `a_i` are fixed (`(a_i)_i = range(1,10)`) and `b` corresponds to the permutations (the lines, cols and small squares). So `b` is commited in `gamma` and I think it is okay :-)

XuyangSong: so the `a_i`(or `cell_integers` in the code) should be in constant column, not advice column?

Acentelles: `gamma` is deterministic but in order to get the right `gamma`, you need to have a solution for the sudoku

taiga_halo2/examples/sudoku/app/valid_sudoku.rs
XuyangSong: sorry, it's already public. I misread `assign_free_instance` as `assign_free_advice`.
You added a new `assign_free_instance`. That's very useful!

bazzilic (approved):
All looks good to me!

PR #93, opened 2022-12-01
Test for `to_affine`, reorganization of curve circuit

simonmasson:
A simple test of JacobianToAffine (used in the hash_to_curve), and a reorganization so that circuits are split into different (smaller) files.


XuyangSong (approved):
lgtm

PR #94, opened 2022-12-09
transaction and partial transaction

XuyangSong:
- [x] Add transaction and partial transaction structs and according APIs
- [x] give an example(test) to create transaction
- [x] params and action circuit keys(proving and verifying) pre-computation
- [x] refactor and refine `ValidityPredicateCircuit` trait
- [x] modify some basic structs and add some useful apis
- [ ] add binding signature to check sum balance 
- [x] `Note` and `Application` structs refactor(move user and vp_data to Note) 
- [ ] generalize `User` to `vp_data_nonhashed` and according vps
- [ ] decide how to handle nullifier_key in `User` if we generalize it


XuyangSong:
Can we review and merge this PR first? Although there are some issues not finished. The binding signature and `vp_data_nonhashed` are independent. When merged, we can parallel some examples(like Sudoku example) and the basic features.

XuyangSong:
`nullifier_key(nk)` is to derive `nullifier` to prevent double-spent.
In general, `nk` is wrapped in user address, the sender can get the recipient's address but can not get the `nk` of the address. When spending, the sender needs to provide the `nk`, prove the validity(struct) of address and derive the `nullifier`.
In concrete, we are using `user_address = Com(Com(send_vp, nk), recv_vp)`. Sender can get recipient's `Com(send_vp, nk)` and `recv_vp` as address info. When spending, the sender will prove her own address format `user_address = Com(Com(send_vp, nk), recv_vp)` in circuit with correct `nk` and then derive nullifier.
We are going to generalize the `user` and make it generic data defined in application vp, we will lose the fixed format of `user` as a result. Therefore, we may reconsider the protocol about `nullifier_key`

taiga_halo2/src/circuit/vp_examples/field_addition.rs
```diff
@@ -69,36 +78,59 @@ impl FieldAdditionValidityPredicateCircuit {
             b,
         }
     }
+}
+
+impl ValidityPredicateInfo for FieldAdditionValidityPredicateCircuit {
+    fn get_spend_notes(&self) -> &[Note; NUM_NOTE] {
```
Acentelles, line 84: Can we have these methods `get_spend_notes` and `get_output_notes` with a default implementation? It may not be immediate but probably there is a way around

taiga_halo2/src/circuit/vp_circuit.rs
```diff
@@ -6,15 +6,42 @@ use crate::{
...
-    plonk::{ConstraintSystem, Error},
+    plonk::{verify_proof, Circuit, ConstraintSystem, Error, SingleVerifier, VerifyingKey},
+    transcript::Blake2bRead,
 };
-use pasta_curves::pallas;
+use pasta_curves::{pallas, vesta};
+
+#[derive(Debug, Clone)]
+pub struct VPVerifyingInfo {
+    pub param_size: u32,
```
joebebel, line 25: we might have to make all VPs the same size at some point - I'm not sure if we need padding or what

XuyangSong: okay, I think `12` seems good for now.

XuyangSong:
TODO: sudoku example doesn't compile now, fix it by using updated `ValidityPredicateCircuit` trait later.

PR #98, opened 2023-01-05
Update sudoku vp

Acentelles:
Adapt the sudoku vp code to the latest state of the codebase.

The latest changes in `ValidityPredicateCircuit` require the VP to be implement the `Circuit` trait from `halo2`.

I think it sensible to decouple to "underlying" circuit (the logic of the sudoku game in this case) from the validity predicate (which also deals with notes, for example).

If so, I can see a pattern emerging
```
impl Circuit<pallas::Base> for SudokuVP {
    type Config = <SudokuCircuit as Circuit<pallas::Base>>::Config;
    type FloorPlanner = <SudokuCircuit as Circuit<pallas::Base>>::FloorPlanner;

    fn without_witnesses(&self) -> Self {
        Self::default()
    }

    fn configure(meta: &mut ConstraintSystem<pallas::Base>) -> Self::Config {
        SudokuCircuit::configure(meta)
    }

    #[allow(non_snake_case)]
    fn synthesize(
        &self,
        config: Self::Config,
        mut layouter: impl Layouter<pallas::Base>,
    ) -> Result<(), plonk::Error> {
        self.sudoku.synthesize(config, layouter)
    }
}
```
It would be nice to hide this boilerplate

### DIFF
--- a/taiga_halo2/examples/sudoku/app/valid_sudoku/vp.rs
+++ b/taiga_halo2/examples/sudoku/app/valid_sudoku/vp.rs
@@ -1,17 +1,23 @@
-use halo2_proofs::plonk::ConstraintSystem;
-use pasta_curves::pallas;
+use halo2_proofs::{
+    circuit::{Layouter},
+    transcript::Blake2bWrite,
+    plonk::{ self, Circuit, ConstraintSystem, Error, keygen_vk, keygen_pk, create_proof}
+};
+use pasta_curves::{pallas, vesta, Fp};
 
 extern crate taiga_halo2;
 use taiga_halo2::{
     circuit::{
         note_circuit::NoteConfig,
-        vp_circuit::{ValidityPredicateCircuit, ValidityPredicateConfig, ValidityPredicateInfo},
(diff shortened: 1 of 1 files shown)


XuyangSong:
For a vp circuit, there are two parts of constraints. One(`basic_constraints`) is to constrain the basic notes struct in the (partial)tx, the other(`custom_constraints`) is to constrain additional logic(e.g. the Sudoku check constraints will be added here).
Anyway, a (vp)circuit requires the `Circuit` trait implementation from `halo2`. To hide the boilerplate, I added the `ValidityPredicateCircuit` trait and a Macro `vp_circuit_impl!`. Once we implement the `ValidityPredicateCircuit` trait and `vp_circuit_impl!` for a vp circuit, we don't need the `Circuit` trait since it's already implemented implicitly. And we had a `FieldAdditionValidityPredicateCircuit` example [here](https://github.com/anoma/taiga/blob/main/taiga_halo2/src/circuit/vp_examples/field_addition.rs)

bazzilic (approved):

bazzilic:
I'll merge this into `vasily/note_data` branch for now, and as to further changes, we can separately discuss those.
@XuyangSong ok?

PR #102, opened 2023-01-10
A preliminary fix for the sudoku circuit that fails on some sudoku puzzles

bazzilic:
Further investigation required but so far it seems that some byte arrays can't be directly reconstructed into a pallas curve field element. If no `u64` group of 8 bytes starts with a `1` bit, it doesn't fail.


Acentelles (approved):
Excellent! Why is it called preliminary? Are we expecting to change it later?

bazzilic:
Yes, I didn't find the root cause of the issue yet, this is a very crude fix.

PR #108, opened 2023-01-17
generalize user

XuyangSong:
- replace `User` with the general `vp_data_nonhashed`
- move nullifier_key out of `User` and add a new `nk_com` in the `Note`(nk_com = Com(nk, zero)). The fixed zero can be replaced by other user-related keys(like note encryption keys) later.
- The `user_address`(user_address = Com( Com(send_vp, nk), recv_vp)) will be removed. Instead, add an inner `address` variable(address = hash(vp_data_nonhashed, nk_com)). And the `address` will go to `NoteCommitment`. Does it make sense that we still use an address here? Otherwise, the `vp_data_nonhashed` and `nk_com` will go to `NoteCommitment` independently, and we need to extend the `SinsemillaCommit` since we add another 255bits.

~~TODO: merge #97 first and rebase this PR~~


Acentelles (approved):
It'd be great to see an example of a use case where a userVP is encoded into `vp_data_nonhashed`

XuyangSong:
> It'd be great to see an example of a use case where a userVP is encoded into `vp_data_nonhashed`

Good suggestion! I'll add it, though we only have the dummy or empty userVPs for now.

taiga_halo2/src/transaction.rs
```diff
@@ -397,7 +398,17 @@ fn test_transaction_creation() {
...
+            trivail_vp_description.clone(),
+            trivail_vp_description.clone(),
+        ];
+        // Encode the app_logic_vps_description into vp_data_nonhashed
+        // The encoding method is flexible and defined in the application vp.
+        // Use poseidon hash to encode the two logic vps here
+        let vp_data_nonhashed = poseidon_hash(
+            app_logic_vps_description[0].get_compressed(),
+            app_logic_vps_description[1].get_compressed(),
+        );
```
bazzilic, line 411: I'm a little confused by this line. We put a _hash_ of the VK into a variable called `nonhashed`.

XuyangSong: Right, maybe we can have a better name?
The hash we used here is to encode the VPs(VKs). We can have other encoding methods.
The `nonhashed` means that the `vp_data_nonhashed` will not be hashed/encoded in value-base

Acentelles: What about:
- `app_vk` (instead of `application_vp`)
- `app_data` (instead of `vp_data`)
- `user_data` (instead of `vp_data_nonhashed`)

Acentelles:
Why don't we store the `user_vp` (or `vp_data_nonhashed` as a `ValidityPredicateDescription` instead of as a field element?

It seems to me that both `app_vp`and `user_vp` are identical in nature (different purposes though). For example, in order to prove ownership of a note, the owner must provide proofs of both the `app_vp` and the `user_vp`.

XuyangSong:
> Why don't we store the `user_vp` (or `vp_data_nonhashed` as a `ValidityPredicateDescription` instead of as a field element?

By my understanding, `vp_data_nonhashed` doesn't have a specific struct. It can be one VP or several user VPs. It also can be any other data that will be encoded to a field element. The application VP is responsible to explain the concrete encoding and meaning of the `vp_data_nonhashed`

taiga_halo2/src/transaction.rs
XuyangSong: I agree with the first two renames.
I'm not sure if we need data, which is not about the user, will be encoded in `vp_data_nonhashed`.

XuyangSong:
From Alberto's suggestion, let's merge it.
The remaining discussion and changes will be in a separate branch.

PR #114, opened 2023-01-23
Update validity predicates and transaction section

Acentelles:


Acentelles:
For @vveiln to pick the bits and pieces she finds useful from this work

PR #115, opened 2023-01-26
Merkle Trees

Acentelles:
- [x] Test Merkle Trees implementation
- [x] Use LR (left-right) struct instead of bool for clarity


XuyangSong (approved):
lgtm

PR #117, opened 2023-01-31
Rename and move is_merkle_checked out of value base derivation

XuyangSong:
- rename `vp_data` to `app_data`
- rename `application_vp` to `app_vk`
- rename `ValidityPredicateDescription` to `ValidityPredicateVerifyingKey`
- use `NoteType` to wrap up the `app_vk` and `app_data` in Note
- rename `vp_data_nonhashed` to `app_data_dynamic `
- move `is_merkle_checked` out of value base derivation. there is a temporary fix because of a possible padding bug in the Poseidon hash gadget in Halo2


Acentelles (approved):
Looks good!

Let's have the `NoteType` and `app_data_dynamic` changes in a separate PR

PR #119, opened 2023-02-01
Move proof to src and remove keys.rs

Acentelles:


XuyangSong (approved):
lgtm
Could we further replace the `create_proof` and use the `Proof` in other files?

PR #122, opened 2023-02-06
Allow using user private keys for signature checks

Acentelles:
Via the `key_dynamic` field

### DIFF
--- a/taiga_halo2/src/circuit/action_circuit.rs
+++ b/taiga_halo2/src/circuit/action_circuit.rs
@@ -153,7 +153,7 @@ impl Circuit<pallas::Base> for ActionCircuit {
             add_chip,
             self.spend_note.clone(),
             ACTION_NF_INSTANCE_ROW_IDX,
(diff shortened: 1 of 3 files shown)


taiga_halo2/src/circuit/action_circuit.rs
```diff
@@ -153,6 +153,7 @@ impl Circuit<pallas::Base> for ActionCircuit {
             add_chip,
             self.spend_note.clone(),
             ACTION_NF_INSTANCE_ROW_IDX,
+            pallas::Base::zero(),
```
XuyangSong, line 156: I found I might need to be corrected about the private key. And we may not need to store the private key in the note.
For signature check vp, we only need the public key to be encoded in `app_data_dynamic`( like app_data_dynamic = encode(sig_check_vp || public_key) ). When generating the proof of signature check vp, we input a valid signature as a witness.
Does it make sense? Sorry about the misleading

Issue #125, opened 2023-02-15
Future development of the Taiga overview


vveiln:
I believe this things might make the Taiga overview more complete:
- binding signature high-level overview
- note encryption high-level overview
- more complex examples of applications
- explain functional privacy achieved with halo2 

This list is very tentative and not all entries might require a full explanation, but it definitely will take time to learn and experiment.

Issue #112, opened 2023-01-23
Update tech spec


vveiln:
- [ ] clean up
- [ ] (maybe) add sections about intent app and token app

PR #126, opened 2023-02-15
Taiga overview big update

vveiln:
Here is the list of things added or removed to the overview (writing from the top of my head):

- removed the outdated examples (new to be added)
- removed the concepts of sendVP, recvVP, userVP from the Taiga-level description
- removed the blinding section (not relevant for Halo2)
- removed the old files and images
- updated the proving system information (PLONK -> Halo2)
- updated the old diagrams 
- update the note structure
- reorganized sections to have more natural order
- extracted the implicit intent application description from the execution model
- described the concept of sub-VP, VP hierarchy

Besides this, I rewrote most of the descriptions to make them more clear.

Things to be added later to the overview (see issue #125) :
- binding signature high-level overview
- note encryption high-level overview
- more complex examples of applications
- explain functional privacy via halo2 accumulation

Things that are not covered in the overview update:
- Taiga specification update is a subject of the issue #112 

I would recommend reviewing the overview by compiling the mdbook:
```
cd taiga/book
mdbook serve --open
```


book/src/app.md
```diff
@@ -1,54 +1,33 @@
...
-`App` define the type of note (e.g. XAN, ETH, BTC). Each application is identified by an address `appAddress` (the same way as user address identifies a user) and has its own VP `AppVP`.
+### Application logic
+Application VP is the defining structure of an application that contains the application logic. 
+Every time the application state is about to change, the application VP is called to validate the state transition.
+If the application VP doesn't consider a state transition valid, the state will not be changed.
 
-### Application VP
-Each application has its own [validity predicate](./validity-predicates.md) `AppVP` that defines the conditions on which the application can be sent/received/etc (e.g. whitelist VP that only allows using the application a specified set of users). As with other VPs, `AppVP` checks that input and output notes of the tx satisfy certain constraints.
-It is required that the `AppVP` of the applications involved in a tx evaluated to `true`.
+An application VP might require validity of other VPs, enforcing the VP hierarchy with the application VP being the main VP checked.
```
Acentelles, line 8: It'd be worth mentioning at some point how the subVPs are checked. 

From previous discussions, this is done by checking the hash of the subVP inside the appVP and enforcing its validity separately.

book/src/validity-predicates.md
```diff
@@ -1,122 +1,61 @@
...
 
-* Field addition/multiplication
-* Elliptic curve addition and scalar multiplication
-* Poseidon hash
+![img.png](images/vp_img.png)
 
-#### Addresses
+Taiga uses the Halo2 proving system and validity predicates are represented as [PLONKish circuits](https://zcash.github.io/halo2/concepts/arithmetization.html). 
+For privacy reasons, all Taiga VPs share the same configuration (the set of building blocks available for the circuits), 
+and distinct VPs are created by specifying the *selectors* (that select which blocks to use).
```
Acentelles, line 40: > For privacy reasons, all Taiga VPs share the same configuration (the set of building blocks available for the circuits), 
> and distinct VPs are created by specifying the *selectors* (that select which blocks to use).

I don't think this is true; I think each VP can have a different configuration.
E.g.
```
#[derive(Clone, Debug)]
pub struct SudokuVPConfig {
    note_config: NoteConfig,
}
```
but we can add any other configurations if needed

book/src/validity-predicates.md
```diff
@@ -1,122 +1,61 @@
...
 
-Informally, transactions take a private subset of unspent notes from the Taiga note set, publicly reveal their nullifiers, and reveal a new set of note commitments to add to the Taiga note set. The Action/Execute circuit verifies consistency of this state transition, but does not check directly its validity. Instead, the validity predicate circuits must check the validity of the state transition. The following VPs are called:
+#### Public Inputs
 
-* The spending VP for every spent note
-* The app VP for every spent and created note
-* The receiving VP for every created note
+* $\{nf_i\}$ - the set of revealed nullifiers in the transaction
+* $\{cm_i\}$ - the set of new note commitments created in the transaction
+* $e$ - the current Taiga epoch (used for time-tracking)
```
Acentelles, line 54: I don't think these are public inputs for VPs but for transactions

book/src/app.md
```diff
@@ -1,54 +1,33 @@
...
 
-In Taiga, VPs are shielded, so instead of showing that `AppVP` evaluates to `true` publicly, a ZK proof is created. To make sure that `AppVP`  evaluates to `true`, an observer can verify the proof (using the verifier key):
+#### Application address
+Each application has an address that identifies the application. The application address is derived from its application VP (or VPs, if there is more than one).
```
Acentelles, line 11: Do we have more than one application VP at the moment? I think it would help to remove "(or VPs, if there is more than one)" since it brings (to me) some questions. Or if we want to talk about how to have multiple applications, we can have a section for it.

Acentelles: Is the address of an application its verifying key?

book/src/app.md
```diff
@@ -1,54 +1,33 @@
...
 
-```verify(AppVP_proof, app_VK) = True```
+### Application state
+As Taiga works in the UTXO model, application states are stored in notes, and changing the application state is done by spending the old application state notes and creating new application state notes.
 
-### Application Address
-Each app is identified by an address that is derived from its verifier key `app_VK`:
-`appAddress = Com(app_VK)`
+The application a note belongs to is indicated by the note's type which is derived using the application address, binding the note with the application.
+One application can have notes of multiple types, all of which are derived using the application address and some additional information that helps to distinguish the application types.
```
Acentelles, line 17: I'm not sure I understand this. Do you mean that a note can access the other 3 notes from which the transaction is executed?

book/src/intro.md
```diff
@@ -0,0 +1,33 @@
+# Introduction
+
+Taiga is a shielded state transition protocol that allows applications built on top of it to enjoy the advantages of fully shielded multi-party state transitions 
+(hiding the application type, the data associated with it, involved parties, etc) without giving up the application complexity.
+
+![img.png](images/Intro_Taiga_Apps.png)
```
Acentelles, line 6: Would it make sense to have the Apps boxes inside the Taiga box? I see Taiga as the underlying protocol behind every app

book/src/notes.md
```diff
@@ -1,87 +1,67 @@
...
 
-```rust
-pub struct Note<CP: CircuitParameters> {
-	pub owner: User<CP>,
-	pub app: App<CP>,
-	pub value: u64,
-...
-}
-```
+As notes are immutable, notes cannot be modified, only spent or created. "Modifying" a note would be done by spending the current note and creating the same, but modified, note.
```
Acentelles, line 5: It makes more sense to me to talk about spending the current note and creating a note or multiple notes of the same type (maintaining the balance)

book/src/notes.md
```diff
@@ -1,87 +1,67 @@
...
+pub app_data_dynamic: pallas::Base,
+pub value: u64,
+pub nk_com: NullifierKeyCom,
+pub is_merkle_checked: bool,
+...
 }
 ```
+#### The application data
+Notes contain quite a lot of data provided by the application the notes belong to. 
+Part of that data is more long-term (e.g. the application VP) and used to derive the note's `value_base`, while
```
Acentelles, line 43: I'd use note's type instead of introducing value_base

book/src/notes.md
```diff
@@ -1,87 +1,67 @@
...
+the other part is more contextual (e.g. the input parameters for the application VP) and is stored in `app_data_dynamic`.
```
Acentelles, line 44: > (e.g. the input parameters for the application VP)
input parameters for the appVP can also be stored in `app_data`

Maybe we can use the example of a hash of the other subvps that will also need to be checked

book/src/notes.md
```diff
@@ -1,87 +1,67 @@
...
+Notes contain quite a lot of data provided by the application the notes belong to. 
+Part of that data is more long-term (e.g. the application VP) and used to derive the note's `value_base`, while
+the other part is more contextual (e.g. the input parameters for the application VP) and is stored in `app_data_dynamic`. 
+In general, all data that is not used to derive the value base is stored in the `app_data_dynamic`.
 
-, where:
-- `data` is additional information that might be useful to describe notes of a certain type (e.g. NFTs)
-- `rho`  and `psi` are the values used to compute the note's nullifier
-- `rcm` is randomness used to compute the commitment `cm`
+#### Note values
```
Acentelles, line 47: Maybe it's worth mentioning that value encodes fungibility

book/src/app.md
vveiln: I'm not sure that this level of detalization aligns with the rest of the overview. I tried to remove the low-level details, and this, I believe, more of an implementation detail. But I agree that we need to mention it, but perhaps in the spec

book/src/validity-predicates.md
vveiln: Ah, it is a piece of the old document and I guessed it wouldn't change. Maybe I'll just remove it as it seems to be a very low-level detail as well

book/src/validity-predicates.md
vveiln: I'm not sure it would make sense for transactions to have public or private inputs. It does make sense for the Action circuit, but we also have it for VPs, at least according to the current spec. The spec is outdated, of course, but it seemed like an intentional decision for both circuits to have nullifiers and commitments as a public input, and it feels like it wouldn't be affected much by what we've done since then, so I'm not sure how it is now. I'll clarify

book/src/app.md
vveiln: The address is derived from the application verifying key, but doesn't equal it. I think mentioning verifying key explicitly is most of the time unnecessary for the overview, this is a detail that doesn't help to understand the architecture more, but adds more information. It would be more appropriate to describe in details in the spec

vveiln: About multiple VPs: to me it was more about available options, but I see how it can be confusing for the reader. I'll remove it

book/src/app.md
vveiln: An application A can have notes of types AA, AB, AC, etc. All of them are binded to the application A, but they have distinct types and don't balance each other. They are independent as if they were notes of different applications, but they are actually binded to the same application. For example, intent notes are like that. The value base of intent notes are derived using userVPs, so they are different for each userVP, but they still belong to the same intent application.

book/src/intro.md
vveiln: I see it the same way and I agree that arrows aren't good, but I'm not sure about the boxes inside one box either. Kinda gives a feeling that the amount of boxes that the Taiga box can fit in is limited. When I think about Taiga applications, the diagrams for ISO/OSI model come to mind, but it is hard to copy it with only two layers. I'll see what I can do, have a couple of ideas

book/src/notes.md
vveiln: Yeah, not the best wording

book/src/notes.md
vveiln: I agree that mentioning value_base is unnecessary, but I think it would be better to change it in the code. To have note structure as
```
Note {
type: ValueBase;
}
```

Makes more sense to me.

book/src/notes.md
vveiln: Input parameters change every time, so storing it in `app_data` i.e. long-term structure that is used to derive the value base doesn't sound right. So, the only field it can be stored in is `app_data_dynamic`

book/src/notes.md
vveiln: I'm not sure how it would help. If it is about fungible vs non-fungible tokens, I'm not sure it is true because non-fungible tokens also need to be balanced and to have a value for that, even though carrying a different meaning. If it isn't about it, it feels misleading bc many people would think about the distinction above.

XuyangSong (approved):

book/src/action.md
```diff
@@ -1,33 +1,23 @@
...
+* the note existed before (can you spend a note that doesn't exist?),
+* the note haven't been spent yet,
+* the spender has a right to spend it (do you own the note?),
+* and that the application the note belongs to approves the action (sure there is some application approves the check, but is it the one the note belongs to?).
 
-## Creating a note
+### Check the new state (created notes)
 
-To create a note, the user also needs to prove the right to do so. `AppVP` must allow the creation of the note, as well as recipient user's `RecvVP`:
+For the notes being created, the Action circuit checks that:
```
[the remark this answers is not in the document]
vveiln: Yes! I'll add it

book/src/performance.md
```diff
@@ -0,0 +1,11 @@
+# Performance
+To give you a sense of performance, we benchmarked the action circuit and a simple VP circuit. 
+Action circuit is always the same and has a size about about $2^{15}$ gates, while the VP circuit size changes as the VPs differ from application to application.
```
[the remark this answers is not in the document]
vveiln: Oh, okay!

book/src/action.md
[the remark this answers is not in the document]
vveiln: But not there probably, I'll mention it as a future feature somewhere, and once we have it I'll update the overview with the VE stuff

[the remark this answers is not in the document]
vveiln: Ah nvm, I'll add it there too 😅

Acentelles (approved):

book/src/app.md
```diff
@@ -1,54 +1,33 @@
...
+If the application VP doesn't consider a state transition valid, the state will not be changed.
 
-### Application VP
-Each application has its own [validity predicate](./validity-predicates.md) `AppVP` that defines the conditions on which the application can be sent/received/etc (e.g. whitelist VP that only allows using the application a specified set of users). As with other VPs, `AppVP` checks that input and output notes of the tx satisfy certain constraints.
-It is required that the `AppVP` of the applications involved in a tx evaluated to `true`.
+An application VP might require validity of other VPs, enforcing the VP hierarchy with the application VP being the main VP checked.
 
-In Taiga, VPs are shielded, so instead of showing that `AppVP` evaluates to `true` publicly, a ZK proof is created. To make sure that `AppVP`  evaluates to `true`, an observer can verify the proof (using the verifier key):
+#### Application address
+Each application has an address that identifies the application. The application address is derived from its application VP.
```
bazzilic, line 11: Should it be explained here what exactly is meant by the word "address"?

book/src/app.md
```diff
@@ -1,54 +1,33 @@
...
-```verify(AppVP_proof, app_VK) = True```
+### Application state
+As Taiga works in the UTXO model, application states are stored in notes, and changing the application state is done by spending the old application state notes and creating new application state notes.
 
-### Application Address
-Each app is identified by an address that is derived from its verifier key `app_VK`:
-`appAddress = Com(app_VK)`
+The application a note belongs to is indicated by the note's type which is derived using the application address, binding the note with the application.
+One application can have notes of multiple types, all of which are derived using the application address and some additional information that helps to distinguish the note types.
+Notes of distinct types are independent of each other, unless explicitly designed.
```
bazzilic, line 18: "unless explicitly designed _otherwise_"?

book/src/app.md
```diff
@@ -1,54 +1,33 @@
...
 
-//app address can be used to create notes of that app;
-let app_address = app.address().unwrap();
-```
-This example is reproducible with [this file](https://github.com/anoma/taiga/blob/main/src/doc_examples/app.rs) or with the command
-```
-cargo test doc_examples::app::test_app_creation
-```
+In some sense, Taiga applications are similar to Ethereum applications, but there are two key distinctions that come to mind:
+* Ethereum uses smart contracts when Taiga uses validity predicates to describe the application logic (learn more about the difference [here](https://github.com/anoma/whitepaper/blob/main/whitepaper.pdf), page 3)
```
bazzilic, line 31: Technically, the whitepaper says that VPs _are_ smart contracts, just designed differently: "validity predicates are an architecture for smart contracts..."

book/src/app.md
```diff
@@ -1,54 +1,33 @@
...
+* Taiga applications are shielded by default, but can be defined separately over the transparent pool as well. The shielded and transparent parts of the application can interact with each other, but whatever happens in Taiga is always shielded
```
bazzilic, line 32: Is that still the case? Considering that unshielded execution is now also part of Taiga?

book/src/deprecated/examples.md
```diff
@@ -0,0 +1,257 @@
...
+    verifier_data.key,
+    &proof,
+    &verifier_data.pi,
+    b"Test",
+)
+.unwrap();
+```
+
+
+#### Creationg a note
```
bazzilic, line 84: Typo: Creating

book/src/action.md
```diff
@@ -1,33 +1,23 @@
...
-To spend a note, the users need to prove that they are allowed to do so, namely:
-- To make sure that the correct `AppVP` is checked, prove the relationship between `app_VK` and the application field in the spent note.
-  ![img_3.png](img/action_img_3.png)
-- To make sure the correct `SendVP`(read more about `SendVP` [here](./users.md)) is checked, prove the relationship between `send_VK` and the owner field in the spent note.
-- Compute the nullifier of the spent note (using the nullifier key `nk`) so that it cannot be double-spent. To make sure that the nullifier is computed correctly, prove the relationship between `nk` and the owner field in the spent note.
+### Check the old state (spent notes)
 
-![img_1.png](img/action_img_1.png)
+For spent notes, the Action circuit checks that:
+* the note existed before (can you spend a note that doesn't exist?),
```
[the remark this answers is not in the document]
vveiln: Oh, I somehow missed this comment!

I feel like it is better define it in a way that we can't spend a note that doesn't exist 😅

I get your point, I think it relates to the same problem of having/not having a local CMtree. Intent notes are dummy anyway, so it is hard to say how much it breaks the description.

But Action circuit does check MT commitment for non-dummy notes, right? The question is if it can make sense to spend a non-dummy intermediate note

book/src/app.md
vveiln: I'm not sure what you mean. I'm referring to an address as an application identifier, so in that sense I'm not sure more explanation is needed

book/src/app.md
vveiln: Well, it is a question of terminology. Smart contract in a broader sense as an idea vs smart contract as what exactly ethereum uses. I'm referring to the latter, the whitepaper refers to the former. I'll rephrase it

PR #127, opened 2023-02-21
sudoku example VPs

XuyangSong:
doc: https://hackmd.io/@yulia/sudoku

- Dealer Intent App VP
1. check `app_data = hash(encoded_puzzle || sudoku_app_vk)`
2. if it is an output note, no constraints
3. if it is a spend note, check that there is a puzzle note with zero value, the state of the puzzle note is equal to  `encoded_puzzle`, and `sudoku_app_vk` is equal to the `app_vk` of the puzzle note.

- Sudoku App VP
1. check the encoding of the current_state
2. check `app_data = hash(init_state || current_state)`
4. check `current_state` is a valid sudoku
5. if it is the init sudoku note(output note), check `init_state = current_state`
6. if it is an output note(the puzzle state is updated from the spend note), check `spend_note_init_state = output_note_init_note` and `spend_note_vk = output_note_vk`
7. if it is a spend note, the `output_note_current_state` is a better solution than the `spend_note_current_state`. If the `output_note_current_state` is the final solution, check the value of output_note is zero, else check the value of output_note is one.


taiga_halo2/examples/sudoku/app/dealer_intent_vp.rs
```diff
@@ -0,0 +1,417 @@
...
+    utils::poseidon_hash,
+    vp_circuit_impl,
+    vp_vk::ValidityPredicateVerifyingKey,
+};
+
+#[derive(Clone, Debug, Default)]
+struct DealerIntentValidityPredicateCircuit {
+    spend_notes: [Note; NUM_NOTE],
+    output_notes: [Note; NUM_NOTE],
+    // The note that vp owned is set at spend_notes[0] or output_notes[0] by default. Make it mandatory later.
```
Acentelles, line 42:
```suggestion
    // TODO: The note that vp owned is set at spend_notes[0] or output_notes[0] by default. Make it mandatory later.
```

taiga_halo2/examples/sudoku/app/dealer_intent_vp.rs
```diff
@@ -0,0 +1,417 @@
...
+    // is_spend_note helps locate the target note in spend_notes and output_notes.
```
Acentelles, line 43: Do you think it's worth sketching how locating the target note happens here in plain text?

taiga_halo2/examples/taiga_sudoku/app_vp.rs
```diff
@@ -0,0 +1,800 @@
...
+        output_note_variables: &[OutputNoteVar],
+    ) -> Result<(), Error> {
+        let is_spend_note = assign_free_advice(
+            layouter.namespace(|| "witness is_spend_note"),
+            config.advices[0],
+            Value::known(self.is_spend_note),
+        )?;
+
+        // publicize is_spend_note and check it outside of circuit.
+        layouter.constrain_instance(is_spend_note.cell(), config.instances, 2 * NUM_NOTE)?;
```
Acentelles, line 509: Why do we do `2 * NUM_NOTE`?

taiga_halo2/examples/taiga_sudoku/app_vp.rs
```diff
@@ -0,0 +1,800 @@
...
+        state_update_config: &StateUpdateConfig,
+        triple_mul_config: &TripleMulConfig,
+        value_check_config: &ValueCheckConfig,
+        is_spend_note: &AssignedCell<pallas::Base, pallas::Base>,
+        pre_state: &[AssignedCell<pallas::Base, pallas::Base>],
+        cur_state: &[AssignedCell<pallas::Base, pallas::Base>],
+        _spend_note: &SpendNoteVar,
+        output_note: &OutputNoteVar,
+    ) -> Result<(), Error> {
+        // check state update: the cur_state is updated from pre_state
```
Acentelles, line 357: I can't see how we are checking the current state is updated from the previous state. `state_update_config.assign_region` is just a few copy constraints

XuyangSong: There is a gate to check state updates: https://github.com/anoma/taiga/blob/8d7c6c975e12128478b48431b5f468c970e126d2/taiga_halo2/examples/taiga_sudoku/gadgets/state_update.rs#L55
It checks that the pre_state_cell is zero or pre_state_cell is equal with cur_state_cell

taiga_halo2/examples/taiga_sudoku/app_vp.rs
```diff
@@ -0,0 +1,800 @@
...
+                [2, 0, 6, 1, 8, 4, 9, 7, 3],
+                [1, 0, 9, 3, 2, 6, 8, 4, 5],
+                [8, 0, 5, 2, 1, 9, 3, 6, 7],
+                [9, 0, 3, 7, 6, 8, 5, 2, 4],
+                [6, 0, 7, 4, 3, 5, 1, 9, 8],
+            ],
+        };
+        spend_notes[0].value_base.app_data =
+            poseidon_hash(encoded_init_state, previous_state.encode());
+        spend_notes[0].value = 1u64;
```
Acentelles, line 706: We seem to be constraining the value of the output note to 0 or 1. Are we constraining that the value of the spend note is 1? What if it is another value, like 6?

XuyangSong: I have the constraint.
https://github.com/anoma/taiga/blob/8d7c6c975e12128478b48431b5f468c970e126d2/taiga_halo2/examples/taiga_sudoku/gadgets/value_check.rs#L51

Acentelles: The tests seem to pass if I set the value of the spend note to 6 while the value of the output note is 1

Acentelles (approved):
LGTM

taiga_halo2/examples/taiga_sudoku/app_vp.rs
XuyangSong: Oh, I see.
I only constrained the output note.
It shouldn't be a problem if the value of any created output notes is constrained to 0 or 1. Because the spend notes are valid output notes created before. Do you think it makes sense? Or do we need to add the value constraint for the spend note?

XuyangSong: I need to think about it further. We don't check the existence of the spend note since its `is_merkle_checked` is false. 
There could be a problem here

XuyangSong: I think you're right. We'd better constrain the value of the spend note to be 1 (the note with 0 value will not be spent). The only risk I can come up with is that the spend note value can be set as arbitrary(like in your tests) if we don't have the constraint, but the total balance check will not pass.
I'll fix it tomorrow.

Issue #52, opened 2022-06-28
Implement VP which allows for a signature check over arbitrary data


simonmasson:
We want to be able to sign notes.
Two options for not leaking information on the owner:
* Use randomized signatures as in Zcash
* Use ZK proofs for the verification.

cwgoes:
Question: do we need to do anything in the Taiga architecture in order to allow for signature checks in VPs? Could be as simple as providing a partial transaction hash to each VP (which could then check a signature over it).

Order of operations: Implement the simple thing first, optimise later if and only if necessary. We could implement something with rerandomizable signatures that moves signature checks to outside the circuit (more efficient) but not yet clear if necessary.

cwgoes:
@simonmasson are you working on this rn?

simonmasson:
No but I think Xuyang integrated binding signatures. Is it what you are talking about? I think it is already merged into main.

simonmasson:
I will remove myself from the assignees*

joebebel:
I think the right thing to do is pass the hash of the partial transaction to the VP, then it can do the signature check.

Of the Schnorr variants, I think EdDSA will fit best. I'm not sure if we can use the re-randomization ability of RedPallas, but otherwise the RedPallas algorithm should work well.

Acentelles:
Closed by #128

PR #128, opened 2023-02-22
Schnorr Signature Circuit

Acentelles:
Implements #52


taiga_halo2/src/circuit/schnorr.rs
```diff
@@ -0,0 +1,398 @@
...
+            )?;
+
+            ScalarVar::from_base(
+                ecc_chip.clone(),
+                layouter.namespace(|| "ScalarVar from_base"),
+                &h,
+            )?
+        };
+
+        let R = NonIdentityPoint::new(
```
XuyangSong, line 286: The `R` should be the same as the above `r`. And if we assign(NonIdentityPoint::new) it twice, there might be some problems.

PR #133, opened 2023-03-06
remove the note_data

XuyangSong:
> To keep notes lightweight, we aim to encode any arbitrary data inside app_data and app_data_dynamic and the encoding mechanisms are application dependent, so we don't need note_data.

### DIFF
--- a/taiga_halo2/examples/simple_sudoku/circuit.rs
+++ b/taiga_halo2/examples/simple_sudoku/circuit.rs
@@ -383,15 +383,11 @@ impl plonk::Circuit<pallas::Base> for SudokuCircuit {
 
 #[cfg(test)]
 mod tests {
+    use crate::circuit::SudokuCircuit;
     use halo2_proofs::{arithmetic::FieldExt, dev::MockProver};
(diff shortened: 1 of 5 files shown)


Acentelles:
Can you also update the examples?

Acentelles:
For future reference, it's worth noting the reason of this PR. I believe that we try to keep notes lightweight and we aim to encode any arbitrary data inside `app_data` and `app_data_dynamic` and the encoding mechanisms are application dependent, so we don't need `note_data`. Is this correct @XuyangSong ?

bazzilic:
If I remember our discussion on this correctly, we model two layers of "indirection": notes are stored separately, off chain; and to keep them lightweight, whatever data is attached to the note, that is _also_ stored separately, and the note only keeps some kind of reference to that data. Does my memory serve me right? If so, where is that reference kept in the note?

Acentelles (approved):

Acentelles:
I think that the reference for the data attached to the note that you mention is what is stored in `app_data` and `app_data_dynamic`

PR #139, opened 2023-03-08
check the consistance of nullifier and note_commitment

### LINKED ISSUE #120: Check consistency of nf and cm for VPs in PartialTransaction::verify
We must guarantee that all VPs in the partial transaction take the same notes from the action circuits as inputs.

We can explicitly check the consistency of nf for spend_note and cm for output_note since nf and cm are public inputs in the circuits.

XuyangSong:
To fix #120

### DIFF
--- a/taiga_halo2/src/circuit/vp_circuit.rs
+++ b/taiga_halo2/src/circuit/vp_circuit.rs
@@ -32,6 +32,14 @@ impl VPVerifyingInfo {
         let params = SETUP_PARAMS_MAP.get(&VP_CIRCUIT_PARAMS_SIZE).unwrap();
         self.proof.verify(&self.vk, params, &[&self.instance])
     }
+
+    pub fn get_nullifiers(&self) -> [pallas::Base; NUM_NOTE] {
+        [self.instance[0], self.instance[2]]
+    }
+
+    pub fn get_note_commitments(&self) -> [pallas::Base; NUM_NOTE] {
+        [self.instance[1], self.instance[3]]
+    }
 }
 
 pub trait ValidityPredicateConfig {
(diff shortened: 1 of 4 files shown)


taiga_halo2/src/circuit/vp_circuit.rs
```diff
@@ -32,6 +32,14 @@ impl VPVerifyingInfo {
         let params = SETUP_PARAMS_MAP.get(&VP_CIRCUIT_PARAMS_SIZE).unwrap();
         self.proof.verify(&self.vk, params, &[&self.instance])
     }
+
+    pub fn get_nullifiers(&self) -> [pallas::Base; NUM_NOTE] {
+        [self.instance[0], self.instance[2]]
```
Acentelles, line 37: It looks a bit clunky. What's the logic behind this indexes? Would it be possible to make it more declarative? Can it go wrong (e.g. by instances not having at least for elements or by having elements in the wrong position)?

taiga_halo2/src/transaction.rs
```diff
@@ -125,14 +134,14 @@ impl Transaction {
         self.partial_txs.push(ptx);
     }
 
-    pub fn get_nullifiers(&self) -> Vec<Nullifier> {
+    pub fn get_nullifiers(&self) -> Vec<pallas::Base> {
```
Acentelles, line 137: Would it be possible to keep the types of Nullifier and NoteCommitment?

Acentelles (approved):
Looks good to me mod some comments

taiga_halo2/src/circuit/vp_circuit.rs
XuyangSong: I'll add the constant index names to make it more declarative

PR #143, opened 2023-03-16
look up the owned note variable and get the is_spend_note_flag in the VP circuit

### LINKED ISSUE #138: Generalize the method/API to locate the owned note from the four notes in the VP circuit
The VP takes four notes(two spend notes and two output notes) in one partial txn and treats them equally. We need a general method to help the VP search and locate the note the VP owns in the circuit. 

Sudoku Application uses a temporary solution that is not general but has fewer constraints. The new available proposal:

We can use the nullifier(note commitment) of the spend(output) note as the target tag since the nullifier and the note commitment are unique and the public inputs of proof. And we need to witness and publicize the nullifier(note commitment) of the owned note in the VP circuit as the source key to search in the four target tags and return the self-owned note parameters. Besides, we also need to check the nullifier(note commitment) out of the circuit.

~~The primary feature/method will be added in the `basic_constraints` of the `ValidityPredicateCircuit trait`~~
The `get_owned_note_variable` and the `get_is_spend_note_flag` circuits might not be used in every VP circuit. We provide the essential ability in `custom_constraints,` which can be used when needed.

XuyangSong:
close #138


Acentelles:
Can you explain the overall mechanism/algorithm here or as a comment in the code? (the more details the better)

XuyangSong:
> Can you explain the overall mechanism/algorithm here or as a comment in the code? (the more details the better)

#138

Acentelles:
I saw that. Do you think it is enough?

XuyangSong:
> Do you think it is enough?

Yes.
I've described why we need this feature and the core idea to implement the circuit in #138. And I remember we had a discussion about it in the meeting.
Okay, I can add more details. The mechanism/algorithm is quite simple. Use self-owned note nf/cm as the key, then search in the table, which includes the two spent note nfs, two output note cms, and the corresponding target variables, and finally return the target variable if the self-owned note nf/cm is equal to the target nf/cm.

Let me know what your questions are.

PR #157, opened 2023-04-23
Integrate Vamp-IR into Taiga

XuyangSong:
- [x] Create predicate from vamp_ir circuit
- [x] Add some tests


XuyangSong:
Some problems needed to discuss with Vampir Team:
- [ ] Vampir lib doesn't compile. 
- [ ] How to handle instances(public inputs)? Vampir hardcoded the instances as empty, which needs to fix.
- [ ] `fn prompt_inputs` is a private function in Vampir. We want to create proofs in Taiga. Could they make it public?
- [ ] Extend the range of parameters, which is 2^{0~11} in Vampir. Our VP uses at least 2^12.
- [ ] Could Vampir separate the parameter and circuit serialization? In Taiga, we have the Halo2 parameter in memory and only want the Vampir circuit. Reading the parameter every time is not efficient, since the parameter is huge.

lopeetall:
@XuyangSong can you explain the problem with parameter range in Vamp-IR? The code I'm looking at appears to size the parameters according to the size of the circuit, and we've definitely used circuits with > 2^12 constraints, so I'm surprised at this one

XuyangSong:
> @XuyangSong can you explain the problem with parameter range in Vamp-IR? The code I'm looking at appears to size the parameters according to the size of the circuit, and we've definitely used circuits with > 2^12 constraints, so I'm surprised at this one

Sorry, I misunderstood the parameter `AllowedEnumVariants`. Thanks for the explanation. I removed this one.

bazzilic (approved):
👍

PR #158, opened 2023-04-24
transaction refactor

XuyangSong:
- [x] Refactor transactions to support both shielded and transparent ptx
- [x] Move binding signature in Transaction to check the balance of `ShieldedPartialTxBundle` and `TransparentPartialTxBundle`
- [x] Add an `Executable` trait to unify the interfaces of shielded and transparent ptx
- [ ] Complete the transparent ptx


taiga_halo2/src/transaction.rs
```diff
@@ -201,451 +125,207 @@ impl Transaction {
...
-        &self,
-    ) -> Result<(Vec<Nullifier>, Vec<NoteCommitment>, Vec<pallas::Base>), TransactionError> {
-        // Verify proofs
-        self.verify_proofs()?;
+        BindingVerificationKey::from(vk)
+    }
 
-        // Verify binding signature
-        self.verify_binding_sig()?;
+    fn digest(&self) -> [u8; 32] {
```
joebebel, line 147: If the signature is on a hash of a bundle, then we can't combine differently-signed partial transactions into one bundle.
We probably need to think more about this

joebebel (approved):
Looks good. I think we need to figure out how digest works with partial tx/tx bundles.

taiga_halo2/src/transparent_ptx.rs
```diff
@@ -0,0 +1,62 @@
+use crate::{
+    error::TransactionError, executable::Executable, note::NoteCommitment, nullifier::Nullifier,
+    value_commitment::ValueCommitment,
+};
+use pasta_curves::pallas;
+
+#[derive(Debug, Clone)]
+pub struct TransparentPartialTransaction {
```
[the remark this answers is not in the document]
XuyangSong: It is still being decided.
Most designs about the transparent partial tx are not nailed down in this PR, except for some interfaces shared with shielded ptx.
I'll list all the questions of transparent partial tx/ execution. We can talk about it in the meeting later.

taiga_halo2/src/transaction.rs
XuyangSong: What do you mean by differently-signed partial transactions?
I wanted to digest the whole tx(including all the transparent and shielded ptxs) and do the binding signature for the balance checking. It's not for authenticity and integrity. The digest contains the notes' nf, cm, value_commitments, and anchor. But I might miss some requisite parameters.

XuyangSong: Yes
The binding signature is for the final transaction. We can compute/recompute the binding signature whenever the transaction is finalized.

Issue #160, opened 2023-05-08
Taiga transaction examples


XuyangSong:
Create some transaction examples to explain how Taiga works
- [ ] Basic swap transaction examples: two-party and multi-party token swap without intent
- [ ] Basic intent transaction examples: how intent is created and consumed.
- [ ] transaction that makes multiple partial transactions atomic to handle the note number exceeding the `NUM_NOTE`(two inputs, two outputs) in one partial transaction.

PR #170, opened 2023-06-01
Taiga transaction examples 

XuyangSong:
From #160 

- [x] token vp
- [x] signature verification vp to authorize the token spending
- [x] Basic swap transaction examples: two-party and multi-party token swap without intent
- [x] Basic intent transaction examples: how intent is created and consumed.
- [x] complex intent with two conditions
- [x] partial fulfillment intent
- [x] transaction that makes multiple partial transactions atomic to handle the note number exceeding the `NUM_NOTE`(two inputs, two outputs) in one partial transaction.
- [ ] add note encryption to the examples


XuyangSong:
It's fine to constrain the `is_merkle_checked` flag to the constant in the vp circuit, it will expose the flag though. When we complete the function privacy(verifier circuit), the constant column of the vp circuit will be hidden.

XuyangSong:
The note encryption error was fixed. Add the note encryption to the transaction example and decide what data should go to the encryption of the vp/intent circuit.

XuyangSong:
The receiver_vp that contains the note encryption constraints will be in a new PR. There are still some concrete implementation issues to be figured out.
I'll merge this PR first.

Issue #151, opened 2023-03-24
Integrate Vamp-IR into Taiga


bazzilic:
- [x] Import trivial VPs (`1=1` and `1=0`) and attempt to prove them within taiga codebase; adjust taiga if necessary.
- [x] List questions for Vamp-IR team and functional requirements for Vamp-IR to be integrated into Taiga (e.g., I think currently Vamp-IR is only available as a separate binary, but we’d probably like to see it as a library we can import into taiga code; we need vampir to output a VK, which I am not sure it currently does)
- [x] Further determine scope of this task and add more items below

bazzilic:
Also blocked by https://github.com/anoma/vamp-ir/issues/57

XuyangSong:
Some problems needed to discuss with Vampir Team:
https://github.com/anoma/taiga/pull/157#issuecomment-1519080215

The basic underlying gadgets we need to create Taiga Predicate(mandatory/basic constraints):
Poseidon hash, Elliptic curve operation(addition and multiplication), sinsemilla commitment

XuyangSong:
> Remove the `params` from the `VampIRValidityPredicateCircuit` when vamp_ir supports setting the params size manually.
Some VampIR VP circuit tests failed because of the small `params`. It will be solved by setting bigger `params` manually. We need the feature in VampIR as above.

paulcadman:
Proposed prove and verify vamp-ir API that exports the verifying key https://github.com/anoma/vamp-ir/pull/120

PR #179, opened 2023-06-27
Construct VampIRValidityPredicateCircuit using a vamp-ir file

paulcadman:
This is a small change to the `VampIRValidityPredicateCircuit` implementation to construct a circuit directly from a vamp-ir file.

Previously VampIRValidityPredicateCircuit was constucted from a Halo2 circuit
file that was precompiled by vamp-ir. We can use the vamp-ir `compile` API
to perform this step in directly in taiga instead.

This change only affects the POC `test_create_vp_from_vamp_ir_file` test.

Relates to:
https://github.com/anoma/taiga/pull/157
https://github.com/anoma/taiga/issues/151


XuyangSong (approved):
lgtm

PR #180, opened 2023-06-28
Add API to build a circuit from vamp-ir source 

paulcadman:
There is only vamp-ir API for creating circuits inputs from a JSON file. Taiga will
need to create inputs from note data stored in partial transactions. So
we'll need API in vamp-ir to do this.

This PR adds API on `VampIRValidityPredicateCircuit`, for compiling a vamp-ir file from source.

```
from_vamp_ir_source(
        vamp_ir_source: &str,
        named_field_assignments: HashMap<String, Fp>,
    ) -> Result<Self, VampIRCircuitError>
```

In `vamp-ir_utils` I've also added API missing from the vamp-ir library which we can add to vamp-ir later.

```
get_circuit_assignments(
    module: &Module,
    named_assignments: &HashMap<String, Fp>,
) -> Result<HashMap<VariableId, Fp>, VariableAssignmentError>
```

and 

```
parse(unparsed_file: &str) -> Result<Module, String>
```


XuyangSong (approved):
lgtm

PR #182, opened 2023-07-03
A receiver vp example

XuyangSong:
A receiver vp example that contains the note encryption constraints

- Added a receiver vp and used it in the token swap examples
- The note elements are encrypted in the receiver vp and can be recovered/decrypted from the instances/public inputs.

Note: To extract and encrypt the `rcm` which was a `ScalarFixed` type, I adjusted the sinsemilla commitment(`sinsemilla_commitment = sinsemilla_hash_to_point(msg) + rcm * {R}`). Now we use `sinsemilla_hash_to_point` from Halo2 and manually compute the `rcm * {R}` to construct the `rcm` from the Base field instead of the Scalar field.

We don't have a generic note decryption design because of the unfixed indexes ciphertext. I only give a specialized note decryption implementation for the receiver vp.


bazzilic (approved):

Acentelles (approved):

XuyangSong:
I'll merge it to prevent potential conflict with other new PRs.
We might need minor modifications later when we unify the public input layout(number and positions).

PR #183, opened 2023-07-04
(De)serialization of Transaction, PartialTransaction and Note

### LINKED ISSUE #152: (De-)Serialization of Transaction, PartialTransaction, Note
As we need to transfer such objects as transactions, partial transactions, notes, and some others over the network and between different systems, those need to be serialized and deserialized.

Preliminary, the decision is to use Borsh (https://borsh.io/) (de)serializer due to its performance, availability for multiple languages, _consistent_ serialization (important if we need to hash or sign the transferred object), and a reliable dev team behind it.

It might make sense to define separate DTO structs and put them in a separate crate, so another codebase does not necessarily have to depend on Taiga as a whole to be able to work with these serialized objects.

XuyangSong:
close #152 

- Add the (de)serialization of Transaction, PartialTransaction, and Note
- Change the `fn get_output_cms(&self) -> Vec<NoteCommitment>` to `fn get_output_cms(&self) -> Vec<pallas::Base>`  in `Executable` trait since we only have the x coordinator of note_commiment in the public inputs.

Unsolved problem: We must provide the `ConcreteCircuit` when serializing the `VerifyingKey`(the API is from halo2). For now, it only works for the `TrivialValidityPredicateCircuit`, which is hardcoded in the serialization.


bazzilic:
Do you think it's a good idea to use Borsh (or, generally, to use a binary format) for this? This would technically make it impossible to deserialize any of the structures in non-rust code -- yes, Borsh is available for other languages, but as I understand, it would be very hard to deserialize this structures into anything meaningful without the access to Halo2 structures.

Anyway, this definitely is ok for now, for early testnets, but maybe we should also consider JSON.

XuyangSong:
> Do you think it's a good idea to use Borsh (or, generally, to use a binary format) for this? This would technically make it impossible to deserialize any of the structures in non-rust code -- yes, Borsh is available for other languages, but as I understand, it would be very hard to deserialize this structures into anything meaningful without the access to Halo2 structures.
> 
> Anyway, this definitely is ok for now, for early testnets, but maybe we should also consider JSON.

You've listed the reason why we chose the Borsh in #152.
I don't understand why deserializing the binary format is impossible. If we can read bytes and have the structure's layout fixed when it's serialized, It's trivial to deserialize the structure. 
Regarding deserializing into meaningful without concrete structures, I agree JSON format is more readable. But it doesn't make much difference between JSON and binary format. JSON has a "tag" to help describe the object. Binary allocates the object at the specific position of the bytes. Neither can recover the original structure without concrete structures(like halo2 structs).
Besides, we might need to pay the size and performance cost for JSON.

bazzilic:
> You've listed the reason why we chose the Borsh in https://github.com/anoma/taiga/issues/152.

Oh, right, I forgot about that :) @joebebel suggested Borsh and I thought it to be a good idea at the time, but now I have second thoughts. Either way, switching from one to another (or adding an additional serializer) in the future seems feasible.

If we serialize field elements to JSON, we can serialize them as regular big integers in their string representation (`"98231231236..."`), which can be deserialized in any language. Also, in JSON we can extract particular fields (using something like JPath) from the object without having deserialize everything into a properly constructed and typed structure. I now think that Borsh is more appropriate for (de)serializing DTO-like structures, whereas for notes, ptx, txn, etc. introspection and human readability is a virtue.

Performance and size considerations are most likely negligible in this case, this is probably the lightest and fastest part of Taiga :)

Borsh serialization works well for transferring the objects between Taiga instances though, which is what we actually need to do right now, so this is perfect. Discussion on JSON is a future consideration.

PR #189, opened 2023-07-12
Specs updates

Acentelles:
- Renamed `value_base` to `note_type`, as per Paul's feedback. The definition of value base as "represents the note type" was not very informative.
- Defined some terms
- Updated some types
- Distinguished between dummy and ephemeral notes


book/src/spec.md
```diff
@@ -10,33 +10,33 @@ We use Halo2/IPA with [Pasta curves](https://github.com/zcash/pasta) developed b
...
 
 ### 1.2 Elliptic curves
-||Name|Scalar field| Base field|Purpose|Instantiation|
-|-|-|-|-|-|-|
-|$E_I$|Inner curve|$\mathbb{F}_q$|$\mathbb{F}_p$|ECC gadget| [Pallas](https://github.com/zcash/pasta#pallasvesta-supporting-evidence)
-|$E_M$|Main curve|$\mathbb{F}_p$|$\mathbb{F}_q$|Action and VP circuits| Vesta|
-|$E_O$|Outer curve|$\mathbb{F}_q$|$\mathbb{F}_p$|Accumulation circuit| Pallas|
+|Name|Base field| Scalar field|Purpose|Instantiation|
+|-|-|-|-|-|
+|$E_p$|$\mathbb{F}_p$|$\mathbb{F}_q$|ECC gadget, Accumulation circuit| [Pallas](https://github.com/zcash/pasta#pallasvesta-supporting-evidence)
```
vveiln, line 15: I'm not sure if this unification is helpful. Accumulation circuit and ECC gadget are defined over the same circuit only because Halo2 happened to have curve cycle, but they are not the same conceptually. I think having inner curve and outer curve is helpful.

book/src/spec.md
```diff
@@ -10,33 +10,33 @@ We use Halo2/IPA with [Pasta curves](https://github.com/zcash/pasta) developed b
...
-|`app_data_dynamic`||Commitment to the note's extra data|
-|`v`|${0..2^{64} - 1}$|The quantity of fungible value|
-|`cm_nk`||Commitment to the nullifier key that will be used to derive the note's nullifier|
-|`ρ`|$\mathbb{F}_p$|An old nullifier from the same Action description (see Orchard)|
-|`ψ`|$\mathbb{F}_p$|$ψ = PRF^{ψ}(0, rseed, ρ)$|
-|`is_merkle_checked`|bool|Dummy note flag. It indicates whether the note's commitment Merkle path should be checked when spending the note.|
-|`rcm_note`|${0..2^{255} - 1}$|A random commitment trapdoor $rcm\_note = PRF^{\texttt{rcm\_note}}(1, rseed, ρ)$|
+|`note_type`|$E_p$|Identifier of the note's application|
+|`app_data_dynamic`|$\mathbb{F}_p$|Encoding of the application's extra data|
+|`v`|u64|The quantity of fungible value specific to a note type|
```
vveiln, line 34: I'm not sure the value is specific to a note type. The difference between 1 eth and 1 btc is not in the value part (1) but in the type part (btc/eth)

vveiln: Also using `u64` (as in rust type) adds inconsistency as all other types are defined in a mathy way (either by how they are interpreted or what they are). I got the range from the Orchard specification

Acentelles: The _fungibility_ is specific to a note type

vveiln: Hm, maybe we could say it like that? "The quantity of fungible value. The fungibility is specific to a note type"

XuyangSong:
Do we conclude to rename `value` to `quantity`? @vveiln @Acentelles 
If we do, can we update it in this PR?

book/src/spec.md
```diff
@@ -136,9 +136,10 @@ $ce = Encrypt(note, sk)$
 
 Not all of the note fields require to be encrypted (e.g. note commitment), and the encrypted fields may vary depending on the application. To make sure it is flexible enough, the encryption check is performed in VP circuits.
 
-### 2.6 Dummy notes
-In Taiga, note's value doesn't define if the note is dummy or not, unlike some other systems. Dummy notes can have non-zero value and are marked explicitly as dummy by setting `is_merkle_checked = false` meaning that for dummy notes the commitment's Merkle path is not checked when spending the note. Non-zero value dummy notes are handy for carrying additional constraints (e.g. intents) and balancing transactions.
+### 2.6 Ephemeral notes and dummy notes
+A note is _ephemeral_ if it never gets inserted into the append-only note commitment tree. In Taiga, this means that its existence is limited to the gossip network. An example of an ephemeral note is an intent note. A note is marked ephemeral by setting the `is_merkle_checked` flag to `false`. For ephemeral notes the Merkle authentication path is not checked when consuming the note.
```
XuyangSong, line 140: The ephemeral note will be inserted into the tree. When verifying the note on the chain, we can not tell if the note is ephemeral or not.

Acentelles:
@XuyangSong Do you mean changing the name of the field `value` to `quantity` in the whole codebase?

book/src/spec.md
Acentelles: I find "the quantity of fungible value" really confusing. I updated the definition

book/src/spec.md
Acentelles: Most of the specs are conditioned by us using Halo2, this is not an exception. We have no reference to outer or inner curves specifically in any of our documentation

book/src/spec.md
Acentelles: Updated

book/src/spec.md
vveiln: I still find it useful to keep inner and outer curves separately, even if in Halo2 it is sort of the same. Maybe ask @XuyangSong for the third opinion

book/src/spec.md
```diff
@@ -7,9 +7,9 @@
 We use Halo2/IPA with [Pasta curves](https://github.com/zcash/pasta) developed by Zcash to instantiate our proving system.
 
 ### 1.1 Circuits
-Let `C(x; w) ⟶ 0/1` be a circuit with up to `m` gates. The circuit is represented as polynomials over the chosen curve's **scalar field**, following [plonk-ish arithmetization](https://zcash.github.io/halo2/concepts/arithmetization.html). Commitments are generated over the curve's **base field**.
+Let `C(x; w) ⟶ 0/1` be a circuit. An elliptic curve has arithmetic defined within the _scalar field_ (i.e. the prime number of points of the curve). When we formulate circuits, we use this scalar field. The circuit is represented as polynomials over the chosen curve's scalar field, following [plonk-ish arithmetization](https://zcash.github.io/halo2/concepts/arithmetization.html). When we want to _commit_ to these values in the scalar field, we end up with values in the _base field_. When using these committed values, we encounter what is known as _non-native arithmetic_. This is the motivating factor for proposing a _cycle of curves_.
```
vveiln, line 10: What does "has arithmetic defined within scalar field" mean? I think arithmetic is usually defined (although I'm not sure what does it mean to define arithmetic) through manipulating with coordinates which belong to the base field, so this definition is a bit confusing. I'm generally not sure if we need to explain what an elliptic curve's scalar field is here. Maybe we could just give a link to scalar field definition?

Also I think non-native field arithmetic is the motivating factor for using more than one curve, not necessarily cycle of curves. Cycles are helpful when we have more than 1-layer recursion (our  current case) or just generally switch between fields too often, otherwise we could have used the three curves we used in the zk-garage version and be fine

Acentelles: > What does "has arithmetic defined within scalar field" mean?

For your first question, it means that in kP where k is a scalar and P is a point, k is a scalar field. I took that exact phrase from this post, which I think it was very nicely and succinctly summarised https://blog.icme.io/can-we-avoid-cycles-of-curves/

>  I'm generally not sure if we need to explain what an elliptic curve's scalar field is here. 

The specs used to say "Commitments are generated over the curve's **base field**" which doesn't say a commitment of what. I just added a few clarifications and motivation for our specific example. I don't think adding "(i.e. the prime number of points of the curve)" is not a problem.

> Also I think non-native field arithmetic is the motivating factor for using more than one curve

Sure, it is one of the solutions to the problem. But it is _exactly_ the solution that Halo2 takes, which is what we choose here. If we ever move away from Halo2 we can change it (together with most of the specs). I don't see the problem now

vveiln: > For your first question, I wrote "has arithmetic defined within the scalar field". I could have said the arithmetic of the elliptic curve as a group but I found it redundant. I can add it though if it helps.

Yeah I agree it is redundant, I just don't fully understand this way of phrasing

>I don't think adding "(i.e. the prime number of points of the curve)" is a problem.

No, of course it isn't!

> I don't see the problem now

I find it a bit confusing. Maybe we could just elaborate further, e.g. " to solve the non-native field arithmetic problem, we can use another curve whose scalar field is the base field of the main curve. Cycles of curves allow to switch between fields multiple times without the need to introduce new curve each time"

XuyangSong:
> @XuyangSong Do you mean changing the name of the field `value` to `quantity` in the whole codebase?

Yes, I remember we had a proposal to change the name in spec, but I wonder if we have come to a conclusion.

book/src/spec.md
vveiln: I can't see the update. Could you link the commit?

Acentelles: https://github.com/anoma/taiga/pull/189/commits/3eaee3a28dc21b3beac4829614e154ff01554e99

PR #191, opened 2023-07-16
Public structs should have public fields

bazzilic:
Required for working with the structs in external projects.

### DIFF
--- a/taiga_halo2/src/shielded_ptx.rs
+++ b/taiga_halo2/src/shielded_ptx.rs
@@ -17,21 +17,21 @@ use rand::RngCore;
 
 #[derive(Debug, Clone)]
 pub struct ShieldedPartialTransaction {
-    actions: [ActionVerifyingInfo; NUM_NOTE],
-    inputs: [NoteVPVerifyingInfoSet; NUM_NOTE],
-    outputs: [NoteVPVerifyingInfoSet; NUM_NOTE],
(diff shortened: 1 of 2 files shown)


XuyangSong:
It's okay to make them all public. 
But I didn't want to expose the internal fields and details to the external. I prefer to encapsulate the internal fields and implementation and only provide friendly APIs(maybe we need more APIs, add them if needed) unless we have a solid requirement to operate on the internal fields directly. Do you think it makes any sense?
Furthermore, I plan to remove more `public` for the basic internal structs.

bazzilic:
In the solver code, there's a need to access inputs and outputs of shielded txns, so that's what I initially came to fix, but then I noticed there are public structs with all fields private, so I updated that. We can probably instead wrap those in getters/setters as an API to access them, but is it worth it?

PS: My philosophy in general is against private fields :) at least for public data structures

XuyangSong:
On the contrary, I prefer to minimize the access right to the internal fields and make them public only when necessary.
Let's see others' thoughts. Maybe it's only my personal experience.

XuyangSong:
If all fields accesses are needed from the external, that makes sense to make them public.
I'm curious why the solver needs access in shielded ptx. Only proofs and public inputs are contained in the ptx. Solver can not get helpful plaintext from the shielded ptx. I might miss something?

bazzilic:
I'll close this PR for now, I think I don't need this fields to be public atm.

PR #192, opened 2023-07-17
unify vp public inputs

XuyangSong:
- Rename the `instances` to `public_inputs`
- Fix the size(25) and allocation of VP public inputs: mandatory public inputs allocate `public_inputs[0..6]`; custom public inputs allocate `public_inputs[6..11]`; note encryption public inputs allocate  `public_inputs[11..25]`
- Randomize the public input padding: the random padding is not constrained in the circuit.
- Refine the APIs and implementation of public inputs and note encryption
- TODO: the vamp_ir tests failed; The circuits in vamp_ir tests are too small(the row num is less than 25). The parameter overflows after padding the public inputs.  Change to a complex circuit for vamp_ir tests?


bazzilic (approved):
LGTM

bazzilic:
Is the overflow issue a bug in VampIR?

XuyangSong:
> Is the overflow issue a bug in VampIR?

It's not a bug in VampIR. The trivial vp circuits based on VampIR are no longer compatible with the latest VP struct in Taiga.
In this PR, we enforce VPs to have fixed public inputs. But in VampIR vp tests, the circuits are too small.
It will be solved if we can specify the param's size in VampIR.

bazzilic:
> It will be solved if we can specify the param's size in VampIR. We need the feature in VampIR.

As long as the VampIR team is aware of that :)

PR #197, opened 2023-08-07
BEAM Integration

mariari:
This Topic adds a serialization of many types found inside of Taiga to the BEAM (Erlang, Elixir).

We introduce a new dependnecy `rustler` but keep it hidden under a feature flag. This feature flag is off by default but can be turned on by building with `--features=nif`

Doing so allows any BEAM language to talk to the codebase

Each commit of this Topic has documentation, and it is best to review commit by commit, to see the commentary that goes with what changes.

### DIFF
--- a/taiga_halo2/Cargo.toml
+++ b/taiga_halo2/Cargo.toml
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-rustler = "0.29.1"
+rustler = {version = "0.29.1", optional = true}
 rand = "0.8.4"
 lazy_static = "1"
 blake2b_simd = "1"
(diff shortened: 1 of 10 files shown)


mariari:
@XuyangSong we should probably get the CI to work even with conflicts, as that shouldn't supress any CI checking that happens

XuyangSong:
> @XuyangSong we should probably get the CI to work even with conflicts, as that shouldn't supress any CI checking that happens

That makes sense. 
I tried to find a solution to trigger the CI with conflicts. But it seems to be a GitHub limitation. https://github.com/orgs/community/discussions/26304 
Would you happen to have any suggestions? I can trigger the CI when pushing to `main`, though it doesn't solve the conflict problem.

XuyangSong (approved):

mariari:
> 

yeah, we can still a workflow from namada, as it always goes off

PR #201, opened 2023-08-18
simplify ValidityPredicate trait

XuyangSong:
* merge the `ValidityPredicateInfo` trait into `ValidityPredicateCircuit` trait. Can not merge the `ValidityPredicateVerifyingInfo` trait because it causes an "unsafe object" compile error when creating a dynamic object.
* Remove the `ValidityPredicateConfig` trait and use the concrete ValidityPredicateConfig struct instead since we've unified the VP config. TODO: fix the Sudoku example later, which uses the deprecated `ValidityPredicateConfig` trait.
* Rename the confusing `app_vp_verifying_info/app_vp_verifying_info_dynamic` in the Input/OuputNoteProvingInfo to `application_vp/dynamic_vps`
* tried to replace dynamic VP objects to construct partial transactions but failed. I want to keep the flexibility of Taiga VP so that we can support different kinds of circuits(e.g., native halo2 circuit, Vampir circuit, etc.). We could design an upper API that compiles the uniform format(binary or any other intermediate forms) to a concrete VP(circuit) and call/wrap the underlying Taiga API. Like how we did to support Vampir, Vampir compiles the DSL to a uniform and concrete VampirVP(circuit), which implements the `ValidityPredicateCircuit` of Taiga.


taiga_halo2/src/circuit/vp_circuit.rs
```diff
@@ -56,6 +55,8 @@ use vamp_ir::halo2::synth::{make_constant, Halo2Module, PrimeFieldOps};
 use vamp_ir::transform::compile;
 use vamp_ir::util::{read_inputs_from_file, Config};
 
+pub type ValidityPredicate = dyn ValidityPredicateVerifyingInfo;
+
```
mariari, line 59: Yeah this is a fine simplification, seems it is still used everywhere, however this does bring up a question.

taiga_halo2/src/note.rs
```diff
@@ -96,15 +96,15 @@ pub struct RandomSeed([u8; 32]);
 pub struct InputNoteProvingInfo {
     pub note: Note,
     pub merkle_path: MerklePath,
-    app_vp_verifying_info: Box<dyn ValidityPredicateVerifyingInfo>,
-    app_vp_verifying_info_dynamic: Vec<Box<dyn ValidityPredicateVerifyingInfo>>,
+    app_vp_verifying_info: Box<ValidityPredicate>,
+    app_vp_verifying_info_dynamic: Vec<Box<ValidityPredicate>>,
```
mariari: It seems here we are still essentially being reliant upon a `dyn trait`, if we were to expand this alias, it ends up being

`app_vp_verifying_info: Box<dyn ValidityPredicateVerifyingInfo>`
`app_vp_verifying_info_dynamic: Vec,Box<dyn ValidityPredicateVerifyingInfo>>`

meaning that binding to this is still hard if this is exposed as the outside interface

taiga_halo2/src/note.rs
```diff
@@ -96,15 +96,15 @@ pub struct RandomSeed([u8; 32]);
...
 }
 
 #[derive(Clone)]
 pub struct OutputNoteProvingInfo {
     pub note: Note,
-    app_vp_verifying_info: Box<dyn ValidityPredicateVerifyingInfo>,
-    app_vp_verifying_info_dynamic: Vec<Box<dyn ValidityPredicateVerifyingInfo>>,
+    app_vp_verifying_info: Box<ValidityPredicate>,
+    app_vp_verifying_info_dynamic: Vec<Box<ValidityPredicate>>,
```
mariari: Ditto here

taiga_halo2/benches/vp_proof.rs
```diff
@@ -6,7 +6,7 @@ use pasta_curves::pallas;
 use rand::rngs::OsRng;
 use rand::Rng;
 use taiga_halo2::{
-    circuit::{vp_circuit::ValidityPredicateInfo, vp_examples::TrivialValidityPredicateCircuit},
+    circuit::{vp_circuit::ValidityPredicateCircuit, vp_examples::TrivialValidityPredicateCircuit},
```
mariari, line 9: Probably should have changed with the commit that removed it, but good that the bench still wroks.

mariari (commented):
Overall the changes are fine,

It's just that the dyn traits are still there, but this is a good first step

PR #203, opened 2023-08-23
Mariari/simplify vp trait no integration

mariari:
Just here to show what a topic looks like, before it is integrated and merged into main.

Integration never happens on this branch, but would be merged into main as seen in #204.

This work is just #201, but without any integration.


mariari:
Point was made

PR #205, opened 2023-08-23
Remove reddsa from being a patch

mariari:
Patches change how inner packages compile, this makes it so that if anyone uses the project as a library they too have to include the dependency as a patch, since this dep can work as a dependnecy and not a patch, it is moved into a dep

### DIFF
--- a/taiga_halo2/Cargo.toml
+++ b/taiga_halo2/Cargo.toml
@@ -15,7 +15,7 @@ halo2_proofs = {version="0.3", features = ["dev-graph"]}
 bitvec = "1"
 subtle = { version = "2.3", default-features = false }
 dyn-clone = "1.0"
-reddsa = "0.5"
+reddsa = {git = "https://github.com/heliaxdev/reddsa.git", branch = "taiga"}
 vamp-ir = { git = "https://github.com/anoma/vamp-ir.git", rev = "6d401f8a479951727586ef0c44c42edab3139090"}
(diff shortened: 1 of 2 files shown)


XuyangSong (approved):

taiga_halo2/Cargo.toml
```diff
@@ -15,7 +15,7 @@ halo2_proofs = {version="0.3", features = ["dev-graph"]}
 bitvec = "1"
 subtle = { version = "2.3", default-features = false }
 dyn-clone = "1.0"
-reddsa = "0.5"
+reddsa = {version = "0.5", git = "https://github.com/heliaxdev/reddsa.git", branch = "taiga"}
```
bazzilic, line 18: I'm pretty sure if you do `{ git = ..., branch = }`, then `version=...` is ignored, as it just pulls the HEAD of the branch as the dependency.

XuyangSong: So we'd better remove the `version =`? It seems useless if we use the branch directly?

bazzilic: Yes, I think so. Should be just
`reddsa = { git = "https://github.com/heliaxdev/reddsa.git", branch = "taiga" }`

mariari: I'll open some more topics then, as some of the other dependencies have two specifications of what the revision should be

PR #208, opened 2023-08-25
Bump the rust toolchain version

mariari:
After running cargo update, the following change was added

updating clap_lex v0.5.0 -> v0.5.1

This made it so that the project failed to build with the following error

error: package `clap_builder v4.4.0` cannot be built because it requires rustc 1.70.0 or newer, while the currently active rustc version is 1.67.1
Either upgrade to rustc 1.70.0 or newer, or use
cargo update -p clap_builder@4.4.0 --precise ver
where `ver` is the latest version of `clap_builder` supporting rustc 1.67.1

So I've bumped the version


bazzilic:
Do we even need to restrict the toolchain version? I think we can just use latest.

bazzilic:
I tried, and Taiga seems to build fine if you just remove the fixed version.
I assume, dependencies are built with whatever version they pin? As long as the binary interfaces, marshalling rules, and calling conventions don't change, it should work even if built with different compiler versions, I think; and those probably don't change that often.
If we want to keep using a fixed version, we can at least fix it at the current latest stable.

mariari:
> I tried, and Taiga seems to build fine if you just remove the fixed version. I assume, dependencies are built with whatever version they pin? As long as the binary interfaces, marshalling rules, and calling conventions don't change, it should work even if built with different compiler versions, I think; and those probably don't change that often. If we want to keep using a fixed version, we can at least fix it at the current latest stable.

The code does seem to build fine for `1.72.0` I will bump it up to that. With that said, rust does not use the version they pin. Rust uses the current pinned version, but it would use `1.72.0` to build all the dependencies. This is why the build failed previously, as a new feature was being used.

In the past I've struggled getting a matching version for risc0 and vampir to play with as they require almost mutually exclsuive versions to compile with. Rust does not have a stable ABI, so one can't simply use different versions and link them in necessarily

XuyangSong (approved):

PR #209, opened 2023-08-28
Serialization tweaks

bazzilic:
- Added serde serialization for all relevant structs
- Moved both serde and borsh serialization behind respective feature flags

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,6 @@ members = [
 # ark-serialize = { git="https://github.com/simonmasson/algebra", rev="e2ea75c" }
 # ark-ff = { git="https://github.com/simonmasson/algebra", rev="e2ea75c" }
 # ark-ec = { git="https://github.com/simonmasson/algebra", rev="e2ea75c" }
-halo2_gadgets = { git="https://github.com/heliaxdev/halo2", branch="taiga"}
-halo2_proofs = { git="https://github.com/heliaxdev/halo2", branch="taiga"}
-pasta_curves = { git="https://github.com/heliaxdev/pasta_curves", branch="taiga"}
+halo2_gadgets = { git = "https://github.com/heliaxdev/halo2", branch = "taiga" }
+halo2_proofs = { git = "https://github.com/heliaxdev/halo2", branch = "taiga" }
+pasta_curves = { git = "https://github.com/heliaxdev/pasta_curves", branch = "taiga" }
--- a/taiga_halo2/src/transaction.rs
+++ b/taiga_halo2/src/transaction.rs
@@ -401,9 +401,8 @@ pub mod testing {
         (bundle, r_vec)
     }
 
-    #[cfg(feature = "borsh")]
     #[test]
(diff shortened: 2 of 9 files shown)


taiga_halo2/src/transaction.rs
```diff
@@ -336,8 +349,9 @@ pub mod testing {
         (bundle, r_vec)
     }
 
+    #[cfg(feature = "borsh")]
     #[test]
-    fn test_halo2_transaction() {
+    fn test_halo2_transaction_borsh_serialize() {
```
XuyangSong, line 354: We probably could separate the transaction test and the borsh test as we've added the feature flag?

XuyangSong (approved):

taiga_halo2/src/transaction.rs
bazzilic: Ah, I didn't realize this was a combined test. Yeah, I can separate them. CI runs tests with `--all-features` anyway, so all feature tests are run too. But yeah, let me push an updated test.

bazzilic: Updated the tests.

XuyangSong: Is it worth using the exported TX from `test_halo2_transaction` in `test_halo2_transaction_borsh_serialize` to avoid duplicated TX construction since it takes a long time to build a TX?

bazzilic: I don't know how to pass data between tests though.

bazzilic: Actually, fixed that

bazzilic:
Changes that are left in this branch are some minor things that were not included into the Next branch.

Cargo.toml
```diff
@@ -8,6 +8,6 @@ members = [
 # ark-serialize = { git="https://github.com/simonmasson/algebra", rev="e2ea75c" }
 # ark-ff = { git="https://github.com/simonmasson/algebra", rev="e2ea75c" }
 # ark-ec = { git="https://github.com/simonmasson/algebra", rev="e2ea75c" }
-halo2_gadgets = { git="https://github.com/heliaxdev/halo2", branch="taiga"}
-halo2_proofs = { git="https://github.com/heliaxdev/halo2", branch="taiga"}
-pasta_curves = { git="https://github.com/heliaxdev/pasta_curves", branch="taiga"}
+halo2_gadgets = { git = "https://github.com/heliaxdev/halo2", branch = "taiga" }
+halo2_proofs = { git = "https://github.com/heliaxdev/halo2", branch = "taiga" }
+pasta_curves = { git = "https://github.com/heliaxdev/pasta_curves", branch = "taiga" }
```
XuyangSong, line 13: Could we remove these patches as we've specified them in dependencies?

XuyangSong (approved):

Cargo.toml
bazzilic: No, unfortunately we can't. Unlike what we indicate in out dependencies, these patches also affect the whole dependency tree.

PR #210, opened 2023-08-28
Next

mariari:
DO NOT MERGE

This shows an example of integrating some of the topics open currently into main.

If @XuyangSong approves of the suggested branch, we can simply fast-forward head to it.

We include:
- 208
- 207
- 201
- 197

Note we include a variant of 

- #209 

The changes included to #209 are as follows: 

1. remove the fmt commit (this was done by rerolling the topic with passing `-x cargo fmt` to git
2. the topic rebased on #200 as it relies upon that work and not any other work that is in main

If changes are requested on #209 I can do the changes required to keep it up to date


XuyangSong (approved):
looks good. I'll fast-forward to it.

PR #211, opened 2023-09-04
get CI to work with conflicts

XuyangSong:


bazzilic:
I have limited knowledge of Github CI. If this works, then LGTM :)

bazzilic:
We can also add this (https://github.com/taiki-e/cargo-hack#installation) to our testing pipeline so we have full feature testing coverage - at the expense of like 5x the testing time in the CI.

XuyangSong:
> We can also add this (https://github.com/taiki-e/cargo-hack#installation) to our testing pipeline so we have full feature testing coverage - at the expense of like 8x the testing time in the CI.
> 
> This tool automatically discovers all features in the project and then run tests through all combinations of features. In our case we have 3 features, so it will do 8 test runs.

It seems to take much more time. I'm afraid that our DevOps engineer might complain about it since I was told we have limited resources. I was asked to reduce the CI time and frequency of the MASP project which takes more than half an hour to run CI.
We probably can add it later in another PR if It's fine so far.

PR #214, opened 2023-09-13
replace the sinsemillar note commitment with the poseidon commitment

### LINKED ISSUE #123: Improve the inner confusing `address` naming
As described in #108 :
> The `user_address`(user_address = Com( Com(send_vp, nk), recv_vp)) will be removed. Instead, add an inner `address` variable(address = hash(vp_data_nonhashed, nk_com)). And the `address` will go to `NoteCommitment.` Does it make sense that we still use an address here? Otherwise, the `vp_data_nonhashed` and `nk_com` will go to `NoteCommitment` independently, and we need to extend the `SinsemillaCommit` since we add another 255 bits.

Rename the inner `address` or extend the `SinsemillaCommit.`

### LINKED ISSUE #199: Replace the Sinsemilla note commitment  with Poseidon note commitment
Pros
* Reduce the additional assumptions
* The bit decomposition for the Sinsemilla is not needed anymore.
* We don’t need to hack the Sinsemilla commitment to getting the `rcm` conversion when encrypting the `rcm`.

XuyangSong:
close #199 and resolve #123 incidentally

* Use Poseidon hash to commit the note, `note_cm = poseidon_hash(note, rcm)`. 
* As a result, the note commitment will be a field element instead of a point. Adjust the nullifier derivation from `nf =Extract_P([PRF_{nk}(\rho) + \psi \ mod \ q] * K + note_cm)` to `nf = poseidon_hash((nk || \rho || \psi || note_cm)`
* Remove sinsemilla commitment and the note bits decomposition
* Remove the confusing address(`address = hash(app_data_dynamic || app_data_dynamic)`) in the note. Both `app_data_dynamic` and `nk_com` directly go to the note commitment.
 
Potential improvement: There are seven field elements in Poseidon note commitment. If we composite the `is_merkle_checked`(bool) and `value`(u64 but may be extended to u128 later) to one element(`is_merkle_checked * (1<<128) + value`) in the circuit, we can save one round of Poseidon absorb. Is it worth it?


bazzilic (approved):
LGTM

bazzilic:
> we can save one round of Poseidon absorb. Is it worth it?

I think we can take note of this possibility and leave this optimization for the future.

PR #219, opened 2023-09-25
taiga_halo2::circuit::vp_circuit: Reduce code duplication

therealyingtong:
Reduce code duplication in getters for searchable pairs.

Recommended view: split diff


XuyangSong (approved):
lgtm. Thank you!

PR #220, opened 2023-09-28
borsh bump

bazzilic:
Passes all tests with `--feature-powerset`.


bazzilic:
To me this looks like our github CI is pulling old cache from somewhere

XuyangSong (approved):

XuyangSong:
> To me this looks like our github CI is pulling old cache from somewhere

yeah, that's weird.

PR #225, opened 2023-10-09
fix anchor generation

XuyangSong:
The normal note anchor is generated from the path and note leaf, while the dummy(padding or intent) note anchor is from an existing valid root.

### DIFF
--- a/taiga_halo2/src/circuit/vp_examples/token.rs
+++ b/taiga_halo2/src/circuit/vp_examples/token.rs
@@ -381,6 +381,7 @@ pub fn generate_input_token_note_proving_info<R: RngCore>(
     InputNoteProvingInfo::new(
         input_note,
         merkle_path,
+        None,
         Box::new(token_vp),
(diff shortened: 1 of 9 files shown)


taiga_halo2/src/circuit/vp_examples/token.rs
```diff
@@ -381,6 +381,7 @@ pub fn generate_input_token_note_proving_info<R: RngCore>(
     InputNoteProvingInfo::new(
         input_note,
         merkle_path,
+        None,
```
bazzilic, line 384: So of this is `None`, then it is calculated by the `new` constructor?

bazzilic: Maybe the argument for anchor in the `new` method should be called `custom_anchor`, and if no custom anchor is provided then the standard one is calculated. Just to make it a bit easier to understand for the user.

bazzilic (approved):
LGTM

PR #247, opened 2023-11-09
Updates to readme and cosmetic changes to figures

bazzilic:
1) Some parts of our existing readme were quite outdated
2) Two illustrations we had in the readme were more or less the same illustration, so I only left one
3) Since there are time after time attempts to mock Taiga out, and there is rising interest in Taiga from the outside, I think it's appropriate to give a super condensed crash course on Taiga in the README
4) Added an example of a basic transaction in the end, will add some more in the future


XuyangSong (commented):
lgtm
Could you help review the updates? @vveiln @Acentelles @joebebel

README.md
```diff
@@ -1,81 +1,72 @@
...
 
-The state transition is considered to be valid if validity predicates of 
-all involved parties are satisfied, and the Action circuit check passed 
-successfully.
+Taiga transactions consist of valid partial transactions (sometimes referred to as "intents"). Every partial transaction consists of exactly two input notes and two output notes. Output notes are crafted by the creator of the partial transaction. If a transaction gets executed, its output notes are "created": added to the merkle tree. A Partial transaction is considered valid if all programs (VPs) associated with the notes comprising it are valid (result in `true`). If we don't need all 4 note slots in a partial transaction, we can use "padding" notes: they are same as real notes, but have the `is_merkle_checked` flag set to `false`. Padding notes can also be used to describe the user intent by attaching an intent VP to it. Intent VP can describe arbitrarily complex logic of a valid state change: unless it is satisfied, the whole partial transaction would not be valid and therefore can't be part of a valid transaction.
```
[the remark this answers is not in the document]
bazzilic: It's a discussion that happens as long as I remember. I personally don't like the term ephemeral now, even though I suggested it in the first place 😄 at the time, I thought they only exist within a transaction and disappear without a trace, but then it turned out they still get their nullifiers and commitments added to the state for privacy purposes, so they are not as ephemeral as I thought! But if everyone agrees on ephemeral (or ghost 👻 notes!) that's fine.

Here I call them padding because that's what they are for. Ephemeral notes used for padding. I can see how this particular phrasing might seem like I add a new term, so i'll see how I can fix that.

README.md
```diff
@@ -1,81 +1,72 @@
...
-component of 
-Taiga - applications built on top of Taiga use them to express the 
-application rules, 
-and Taiga makes sure that the rules are being followed with the help of 
-the [Action circuit](./book/src/action.md).
+A [Note](./book/src/notes.md) represents an Anoma resource in Taiga. Notes are immutable and have a "denomination" and a non-negative value, among other fields. Every note has several executable programs (often referred to as [Validity Predicates](./book/src/validity-predicates.md) or VPs) associated with it. Input for programs is the notes plus a number of additional public and private inputs to the VPs. Taiga programs result in a state change, which includes a list of destroyed (or "nullified") notes and a list of newly created notes. Any existing note being input into a program gets destroyed. Destroying a note means revealing its nullifier, a secret value bound to the note. Note ownership is determined by knowledge of the nullifier plus arbitrary logic in a VP associated with the note. Creating a note means computing its commitment (a hash), and adding it to a merkle tree. The global state of Anoma is a hashset of note nullifiers and a commitment tree.
```
vveiln, line 13: I would explain in a couple of words what you mean by "denomination", especially since you use " "

vveiln: > Every note has several executable programs

Doesn't have to be several, unless several can mean just one too

vveiln: > Input for programs is the notes

What notes and what is the purpose of those programs?

vveiln: > Taiga programs result in a state change,

Kinda vague enough to count as correct but still gives an incorrect vibe. They validate the proposed state change, don't create it

vveiln: I feel like we should switch to "consume" instead of "destroy" but it is a low priority question

vveiln: > Note ownership is determined by knowledge of the nullifier plus arbitrary logic in a VP associated with the note.

Questionable. The difference is also somewhat vague, but I would say that note ownership = knowledge of the nullifier (although we might want to clarify the difference between the case when the nullifier is published and is known to everyone), but owning a note alone doesn't allow you to induce the state change (in that sense, the state change is "owned" by a VP, because in the end VPs define if the state is acceptable)

README.md
vveiln, line 15: > (sometimes referred to as "intents")

Intents being or not being partial transactions is a fight we should not bring up here. If we want to say something about intents, we should define them properly

vveiln: > added to the merkle tree.

Not just some merkle tree we have, the commitment tree

vveiln: > Padding notes can also be used to describe the user intent by attaching an intent VP to it.

And many other ways we didn't discuss probably. We shouldn't limit the purpose of these notes to two use cases. They could be used to pad or to carry intent vps, but they are not padding notes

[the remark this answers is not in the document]
vveiln: I personally never liked to call them ephemeral, but this is what we call them in the spec right now, so I'd suggest to stick to ephemeral for now. We def shouldn't call them padding notes: they can be used for padding, but also for much more than just padding

README.md
```diff
@@ -1,81 +1,72 @@
...
 
+Taiga transaction is valid if *1)* all contained partial transactions are valid; and *2)* the transaction is balanced: for every note denomination, the sum of values of all input notes of that denomination is equal to the sum of all output notes of the same denomination. A single partial transaction can be a valid transaction if it is valid.
```
vveiln, line 17: > A single partial transaction can be a valid transaction if it is valid

and balanced

vveiln (commented):
Overall lgtm, left a few comments

README.md
bazzilic: To address some of your questions, the purpose of this particular text is to be short and provide an intuitive overview of the whole lifecycle rather than be fully technically correct (we have specs for that).

> They validate the proposed state change, don't create it
Not since transparent execution is part of Taiga too. Transparent VPs can create state changes.

> switch to "consume" instead of "destroy"
Maybe in the specs, but this was a point that I didn't understand for a while in the beginning that "spent" or "consumed" notes essentially get destroyed.

> I would say that note ownership = knowledge of the nullifier
I meant an (optional) user VP, which would provide an additional factor for authorizing spending the note.

README.md
bazzilic: Right, that's a typo 😄

README.md
vveiln: > Not since transparent execution is part of Taiga too. Transparent VPs can create state changes.

Hmm interesting. The whole idea of state change created by a predicate doesn't make sense to me, are transparent predicates still predicates?

> Maybe in the specs, but this was a point that I didn't understand for a while in the beginning that "spent" or "consumed" notes essentially get destroyed.

I get what you mean, but this phrasing also confuses people. They think it literally stops existing, but it just becomes invalid (yet sort of exists in the form of cm/nf being stored). We can say destroyed for now, people are probably confused no matter what word we use 😅

> I meant an (optional) user VP, which would provide an additional factor for authorizing spending the note.

I'm really nitpicking here, we don't have to specify this detail in the overview if it hinders more than it helps 🙂

bazzilic: > are transparent predicates still predicates

We will know when transparent execution spec is finished! :D

> becomes invalid

I can use the word "invalidated" probably

bazzilic:
I think I addressed most of the comments.

XuyangSong (approved):

PR #248, opened 2023-11-16
transparent execution: verify shielded VPs transparently

### LINKED ISSUE #239: Transparent Execution inside of Taiga
We've had some work about Transparent Execution but no actual transparent vp executed. The following tasks should make it work.
- [ ] run shielded VP circuit transparently through the mock::prover method
- [ ] further unify the shielded and transparent execution API and process
- [ ] add some examples by reusing the shield VPs

XuyangSong:
close #239

### DIFF
--- a/taiga_halo2/src/circuit/vp_circuit.rs
+++ b/taiga_halo2/src/circuit/vp_circuit.rs
@@ -25,6 +25,7 @@ use crate::{
         VP_CIRCUIT_OWNED_NOTE_PUB_ID_PUBLIC_INPUT_IDX, VP_CIRCUIT_PARAMS_SIZE,
         VP_CIRCUIT_PUBLIC_INPUT_NUM,
     },
+    error::TransactionError,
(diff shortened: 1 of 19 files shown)


taiga_halo2/src/circuit/vp_circuit.rs
```diff
@@ -776,6 +778,19 @@ macro_rules! vp_verifying_info_impl {
                 }
             }
 
+            fn verify_transparently(
+                &self,
+            ) -> Result<ValidityPredicatePublicInputs, TransactionError> {
+                use halo2_proofs::dev::MockProver;
```
bazzilic, line 784: This executes the circuit transparently?

bazzilic (approved):

taiga_halo2/src/circuit/vp_circuit.rs
XuyangSong: Yes, it is usually used in tests. The `MockProver` checks the constraints transparently instead of creating a proof.

PR #251, opened 2023-11-21
Sync with resource machine

XuyangSong:
According to the Resource Machine Spec, a bunch of parameters and APIs are renamed.

- Rename note to resource
- Rename note commitment to resource commitment
- Rename note encryption to resource encryption
- Rename note type to resource kind
- Rename value to quantity
- Rename value_commitment to delta_commitment, cv_net to delta
- Rename app_vk to logic
- Rename app_data_static to label
- Rename app_data_dynamic to value
- Rename nonce in resource encryption to encrypt_nonce
- Rename rho to nonce
- Rename nullifier key commitment to nullifier public key

Note: `is_merkle_checked`, `rseed`, and `psi` adjusting will be in separated PRs, in which they need more circuit-level modification.


vveiln:
If the field previously known as `NullifierKeyContainer` can still contain both the commitment (nullifier public key) and the nullifier key, we shouldn't rename it

XuyangSong:
> If the field previously known as `NullifierKeyContainer` can still contain both the commitment (nullifier public key) and the nullifier key, we shouldn't rename it

Reverted the `NullifierKeyContainer`, and rename the nullifier key commitment to the nullifier public key

vveiln (approved):

PR #256, opened 2023-12-14
Update spec.md

th3nyst1c:
I corrected a typo (sincemilla -> sinsemilla)


XuyangSong (commented):
Thank you for the correction.
We now use Poseidon to commit notes instead of the Sinsemilla in the implementation. 
Do we have any plans to update/deprecate the out-of-date Taiga Spec? @vveiln

vveiln:
I think it makes sense to add a warning, stash, or remove the specs from this repo.

The spec in the spec repo is also outdated, but I'll update the Taiga spec there when merging the RM report in the specs, and later we can add the updated version here as well, if we want

th3nyst1c:
Okay, Thx. I close pull request

PR #265, opened 2024-08-23
variable resources

### LINKED ISSUE #263: Memory(resources) sharing between compliance circuits and resource logic
It's straightforward to achieve and totally get rid of the limitation of having a fixed number of inputs and outputs in the transparent situation. The following discussion is gonna be in the shielded situation.

The resource logic requires us to load the necessary resources generated from public inputs of compliance proofs and ensure that we don't reveal which resources are being used. Moreover, all resource logics should have the same circuit structure to prevent information leakage. One simple approach is to pass a fixed number of resources within the access scope (partial transaction) to the logic, as currently done in Taiga. However, this method has obvious drawbacks: we need to add dummy resources when the actual number is less than the fixed one, and split it into multiple partial transactions when it exceeds.

Basically, we need an efficient way to share memory (mostly resources) between compliance circuits and resource logic. One idea is to create a Merkle tree based on all the inputs and outputs from compliance in one partial transaction(the minimal atom execution environment). Instead of directly loading all the resources, the resource logic can simply access specific ones by following their respective paths using only the Merkle root as input. This means that even though resource logics can still maintain the same structure with one root (any information could be compressed into the root theoretically) and one self-owned identifier (nullifier of input resource, commitment of output resource). Plus, this approach allows for flexibility in terms of how many resources are involved in each resource logic(partial transaction).

Additionally, we can also use other vector commitments to avoid checking the merkle path for further performance improvement.

XuyangSong:
The implementation of supporting a variable number of resources, close #263 

- [x] Build a merkle tree from full leaves and get the path for the designated leave
- [x] Existence(merkle path) check circuit supporting variable depthes
- [x] Adjust resource logic APIs and the implementation 
- [x] Adjust out of circuit checks in the partial transaction
- [x] Update tests
- [x] Update application examples  
- [ ] Update Taiga specs and README


taiga_halo2/src/circuit/resource_logic_examples/or_relation_intent.rs
```diff
@@ -220,34 +249,48 @@ impl ResourceLogicCircuit for OrRelationIntentResourceLogicCircuit {
...
+                    &desired_resource.resource.value,
                     0,
                     &mut region,
                 )
             },
         )?;
 
+        // check the desired_resource is an output
+        {
+            let zero_constant = assign_free_constant(
```
Acentelles, line 263: Minor suggestion: shall we have a file with zero and one constants? They come up often

taiga_halo2/src/constant.rs
```diff
@@ -67,29 +69,27 @@ pub const RESOURCE_LOGIC_CIRCUIT_PUBLIC_INPUT_NUM: usize =
     RESOURCE_LOGIC_CIRCUIT_MANDATORY_PUBLIC_INPUT_NUM
         + RESOURCE_LOGIC_CIRCUIT_CUSTOM_PUBLIC_INPUT_NUM
         + RESOURCE_LOGIC_CIRCUIT_RESOURCE_ENCRYPTION_PUBLIC_INPUT_NUM;
-pub const RESOURCE_LOGIC_CIRCUIT_MANDATORY_PUBLIC_INPUT_NUM: usize = 9;
+pub const RESOURCE_LOGIC_CIRCUIT_MANDATORY_PUBLIC_INPUT_NUM: usize = 6;
```
Acentelles, line 72: Do we still want this RESOURCE_LOGIC_CIRCUIT_MANDATORY_PUBLIC_INPUT_NUM?

Acentelles: Same for RESOURCE_LOGIC_CIRCUIT_CUSTOM_PUBLIC_INPUT_NUM and RESOURCE_LOGIC_CIRCUIT_RESOURCE_ENCRYPTION_PUBLIC_INPUT_NUM

Acentelles: Are these now an upper bound instead of a mandatory strict number?

taiga_halo2/src/resource_tree.rs
```diff
@@ -0,0 +1,196 @@
...
+            resource,
+            merkle_path,
+        }
+    }
+
+    pub fn get_resource(&self) -> Resource {
+        self.resource
+    }
+
+    // TODO: fix the depth
```
Acentelles, line 36: Does this TODO still holds?

Acentelles (approved):
Looks good to me!

taiga_halo2/src/circuit/integrity.rs
```diff
@@ -316,11 +321,193 @@ pub fn check_output_resource(
...
+    let logic = assign_free_advice(
+        layouter.namespace(|| "witness logic"),
+        advices[0],
+        Value::known(resource.get_logic()),
+    )?;
+
+    // Witness label
+    let label = assign_free_advice(
+        layouter.namespace(|| "witness label"),
+        advices[0],
```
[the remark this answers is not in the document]
XuyangSong: Good layout could lead to a better performance. It's not optimal to use the first column every time, although we only have several such assignments in circuits.
We can analyze the overall layout of every circuit and assign them to proper columns to improve the performance if necessary.

taiga_halo2/src/constant.rs
XuyangSong: yes, we still need the mandatory strict number.
Variable resource number wouldn't affect the number of resource logic public inputs. Because only the root(one element) of the resources go to the public inputs.

taiga_halo2/src/resource_tree.rs
XuyangSong: it's a typo, fixed. thanks

taiga_halo2/examples/tx_examples/token_swap_with_intent.rs
```diff
@@ -172,95 +141,65 @@ pub fn consume_token_intent_ptx<R: RngCore>(
...
-    // padding the zero resources
-    let padding_input_resource = Resource::random_padding_resource(&mut rng);
-    let mut padding_output_resource = Resource::random_padding_resource(&mut rng);
-
     let merkle_path = MerklePath::random(&mut rng, TAIGA_COMMITMENT_TREE_DEPTH);
 
     // Fetch a valid anchor for dummy resources
     let anchor = Anchor::from(pallas::Base::random(&mut rng));
 
     // Create compliance pairs
```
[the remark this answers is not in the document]
XuyangSong: Its a typo, fixed. thanks

